PR #3, opened 2023-11-20
Adaptive bitrate mods

rickcam9000:
Tweaks to the adaptive bitrate messages and auto clear after 10s


eerimoq:
Fixed in 0.178.0?

PR #4, opened 2024-02-07
Improvements to the "Chat is paused" Message

nofuture2077:
* Changed Icon and text color
* Show the number of new messages

### DIFF
--- a/Moblin/Various/Model.swift
+++ b/Moblin/Various/Model.swift
@@ -245,7 +245,8 @@ final class Model: ObservableObject {
     private var obsWebSocket: ObsWebSocket?
     private var chatPostId = 0
     @Published var chatPosts: Deque<ChatPost> = []
-    @Published var pausedChatPosts: Deque<ChatPost> = []
+    private var pausedChatPosts: Deque<ChatPost> = []
+    @Published var pausedChatPostsCount: Int = 0
     private var newChatPosts: Deque<ChatPost> = []
     private var numberOfChatPostsPerTick = 0
     private var chatPostsRatePerSecond = 0.0
@@ -1849,6 +1850,7 @@ final class Model: ObservableObject {
     }
 
     private func updateChat() {
+        pausedChatPostsCount = pausedChatPosts.count
         if chatPaused {
             return
         }
--- a/Moblin/Various/Model.swift
+++ b/Moblin/Various/Model.swift
@@ -245,7 +245,7 @@ final class Model: ObservableObject {
     private var obsWebSocket: ObsWebSocket?
     private var chatPostId = 0
     @Published var chatPosts: Deque<ChatPost> = []
-    private var pausedChatPosts: Deque<ChatPost> = []
+    @Published var pausedChatPosts: Deque<ChatPost> = []
     private var newChatPosts: Deque<ChatPost> = []
     private var numberOfChatPostsPerTick = 0
     private var chatPostsRatePerSecond = 0.0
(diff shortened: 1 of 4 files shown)


Moblin/Various/Model.swift
```diff
@@ -245,7 +245,7 @@ final class Model: ObservableObject {
     private var obsWebSocket: ObsWebSocket?
     private var chatPostId = 0
     @Published var chatPosts: Deque<ChatPost> = []
-    private var pausedChatPosts: Deque<ChatPost> = []
+    @Published var pausedChatPosts: Deque<ChatPost> = []
```
eerimoq, line 248: Making UI updates on every message received when chat is paused will give errors for busy chats (like 10+ messages/second). The 200 ms timer is used to make chat UI updates. I think we should keep this variable private and create a new published variable called `numberOfPausedChatPosts: Int = 0` that we update when the timer fires.

Moblin/View/Stream/StreamOverlayView.swift
```diff
@@ -2,14 +2,16 @@ import SwiftUI
 
 struct ChatWarning: View {
     var message: String
+    var icon: String
+    var iconColor: Color;
```
eerimoq, line 6: This struct is called ChatWarning because it should warnings. We can rename it to ChatInfo or something like that.

Moblin/View/Stream/StreamOverlayView.swift
```diff
@@ -62,9 +64,9 @@ struct StreamOverlayView: View {
                 }
                 .opacity(model.showChatMessages ? 1 : 0)
                 if !model.showChatMessages {
-                    ChatWarning(message: String(localized: "Chat is hidden"))
+                    ChatWarning(message: String(localized: "Chat is hidden"), icon: "exclamationmark.triangle.fill", iconColor: .yellow)
                 } else if model.chatPaused {
-                    ChatWarning(message: String(localized: "Chat is paused"))
+                    ChatWarning(message: String(format: String(localized: "%d new Messages"), model.pausedChatPosts.count), icon: "pause", iconColor: .white)
```
eerimoq, line 69: I think `"{model.pausedChatMessages} new messages"` is cleaner.

Moblin/Various/Model.swift
nofuture2077: great idea. have added a var pausedChatPostsCount

Moblin/View/Stream/StreamOverlayView.swift
nofuture2077: renamed to ChatInfo. I think this will be useful in the future and to duplicate code isn't helpful

Moblin/View/Stream/StreamOverlayView.swift
nofuture2077: Didn't know that syntax. I'm total beginner when it come to swift. Thanks

Moblin/Various/Model.swift
```diff
@@ -1850,6 +1851,7 @@ final class Model: ObservableObject {
     }
 
     private func updateChat() {
+        pausedChatPostsCount = pausedChatPosts.count
```
eerimoq, line 1854: Move it into the if-statement below, so it is only updated when chat is paused.

nofuture2077: This might introduce bugs when the chat is paused the second time as the counter is not reset to zero. So you would have to reset this to 0 with another assignment.

nofuture2077: so for the first 200ms in second pause the count would be incorrect if I use your method

eerimoq: hmm true, let's keep it like this now and we can improve it later

PR #26, opened 2024-05-13
Refactor Mic

ErAzOr2k:
Mic settings are now saved to database even if RTMP audio is selected

### DIFF
--- a/Moblin/HaishinKit/Media/AudioUnit.swift
+++ b/Moblin/HaishinKit/Media/AudioUnit.swift
@@ -61,7 +61,7 @@ final class AudioUnit: NSObject {
         }
     }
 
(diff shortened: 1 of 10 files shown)


eerimoq (commented):
Started testing =)

After installation a toast showing "Mic not found" is shown when starting the app. When upgrading the app users expects their previously selected mic to still be selected.

The view that opens when tapping the mic selection quick button no long closes automatically when selecting mic. I think it still should do that as most of the time the user will tap the correct mic, and don't want to tap close as well.

Starting Moblin with an external mic plugged it does not automatically select it. Plugging in the external mic with Moblin running does. Not sure how to best handle this.

PR #27, opened 2024-05-16
Refactor AudioUnit

ErAzOr2k:
Just little refactoring. Primary duplicate logic removed.

### DIFF
--- a/Moblin/HaishinKit/Media/AudioUnit.swift
+++ b/Moblin/HaishinKit/Media/AudioUnit.swift
@@ -174,8 +174,7 @@ final class AudioUnit: NSObject {
             return
         }
         // Workaround for audio drift on iPhone 15 Pro Max running iOS 17. Probably issue on more models.
-        let presentationTimeStamp = sampleBuffer.presentationTimeStamp
-       // let presentationTimeStamp = syncTimeToVideo(mixer: mixer, sampleBuffer: sampleBuffer)
+        let presentationTimeStamp = syncTimeToVideo(mixer: mixer, sampleBuffer: sampleBuffer)
         guard mixer.useSampleBuffer(presentationTimeStamp, mediaType: AVMediaType.audio) else {
             return
         }
@@ -212,7 +211,7 @@ extension AudioUnit: AVCaptureAudioDataOutputSampleBufferDelegate {
             numberOfAudioChannels: numberOfAudioChannels
         )
     }
-    
+
     private func syncTimeToVideo(mixer: Mixer, sampleBuffer: CMSampleBuffer) -> CMTime {
         var presentationTimeStamp = sampleBuffer.presentationTimeStamp
         if #available(iOS 16.0, *) {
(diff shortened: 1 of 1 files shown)


Moblin/HaishinKit/Media/AudioUnit.swift
```diff
@@ -175,6 +169,35 @@ final class AudioUnit: NSObject {
     func removeReplaceAudioInner(cameraId: UUID) {
         replaceAudios.removeValue(forKey: cameraId)
     }
+
+    func prepareSampleBuffer(sampleBuffer: CMSampleBuffer, audioLevel: Float, numberOfAudioChannels: Int) {
+        // Workaround for audio drift on iPhone 15 Pro Max running iOS 17. Probably issue on more models.
+        let presentationTimeStamp = syncTimeToVideo(mixer: mixer!, sampleBuffer: sampleBuffer)
```
eerimoq, line 175: `mixer!` means the app will crash if `mixer` is `nil`. I mean, it can probably never be `nil` here, but I think it's a good habit to minimize the use of `!` to reduce the risk of app crashes.

The big exception to this is for settings (the database). There everything is expected to exist, at least in the vast majority of cases.

Moblin/HaishinKit/Media/AudioUnit.swift
```diff
@@ -133,26 +133,20 @@ final class AudioUnit: NSObject {
...
         }
-        let presentationTimeStamp = syncTimeToVideo(mixer: mixer, sampleBuffer: sampleBuffer)
-        guard mixer.useSampleBuffer(presentationTimeStamp, mediaType: AVMediaType.audio) else {
-            return
-        }
-        mixer.delegate?.mixer(
-            audioLevel: .infinity,
-            numberOfAudioChannels: sampleBuffer.formatDescription?.audioChannelLayout?.numberOfChannels ?? 0,
-            presentationTimestamp: presentationTimeStamp.seconds
+        var audioLevel: Float = .infinity
```
eerimoq, line 143: This variable doesn't add any value imo. Just pass `.infinity` to the function below.

Moblin/HaishinKit/Media/AudioUnit.swift
```diff
@@ -175,6 +169,39 @@ final class AudioUnit: NSObject {
...
+        }
+        mixer.delegate?.mixer(
+            audioLevel: audioLevel,
+            numberOfAudioChannels: numberOfAudioChannels,
+            presentationTimestamp: presentationTimeStamp.seconds
+        )
+        appendSampleBuffer(sampleBuffer, presentationTimeStamp, isFirstAfterAttach: false)
+    }
+
+    private func syncTimeToVideo(mixer: Mixer, sampleBuffer: CMSampleBuffer) -> CMTime {
```
eerimoq, line 191: I prefer leaving this as a normal function on file level unless it needs something in the class.

PR #29, opened 2024-05-18
RTMP audio is only visible in the microphone settings when RTMP is connected

ErAzOr2k:
RTMP audio is only visible in the microphone settings when RTMP is connected.
When RTMP is disconnected, the previously selected microphone is selected

### DIFF
--- a/Moblin/Various/Model.swift
+++ b/Moblin/Various/Model.swift
@@ -1198,7 +1198,6 @@ final class Model: NSObject, ObservableObject {
             if self.database.debug!.enableRtmpAudio! {
                 self.media.addRtmpAudio(cameraId: stream.id)
             }
-            stream.isConnected = true
         }
     }
 
@@ -1213,7 +1212,6 @@ final class Model: NSObject, ObservableObject {
             if self.database.debug!.enableRtmpAudio! {
                 self.media.removeRtmpAudio(cameraId: stream.id)
             }
-            stream.isConnected = false
             self.selectMic(mic: self.previousMic)
         }
     }
@@ -5081,7 +5079,7 @@ extension Model {
         if database.debug!.enableRtmpAudio! {
             for rtmpCamera in rtmpCameras() {
                 let stream = getRtmpStream(camera: rtmpCamera)
-                if stream!.isConnected {
+                if isRtmpStreamConnected(streamKey: stream!.streamKey) {
                     mics.append(Mic(name: rtmpCamera, inputUid: rtmpCamera, builtInOrientation: .rtmp))
                 }
             }
--- a/Moblin/Various/Model.swift
+++ b/Moblin/Various/Model.swift
@@ -247,7 +247,8 @@ final class Model: NSObject, ObservableObject {
     @Published var interactiveChat = false
     @Published var blackScreen = false
     private var streaming = false
-    @Published var mic = noMic
+    @Published var currentMic = noMic
+    private var previousMic = noMic
     private var micChange = noMic
     private var streamStartDate: Date?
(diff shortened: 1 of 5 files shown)


Moblin/Various/Model.swift
```diff
@@ -1197,20 +1198,23 @@ final class Model: NSObject, ObservableObject {
             if self.database.debug!.enableRtmpAudio! {
                 self.media.addRtmpAudio(cameraId: stream.id)
             }
+            stream.isConnected = true
```
eerimoq, line 1201: Hmm, is it really necesary to keep track of the RTMP stream status in the stream "settings". I mean, isn't it possible to call `isRtmpStreamConnected()` on the model somehow?

I have not looked into the details though, but prefer keeping the connection state in one place.

ErAzOr2k: cool, I didn't discover the function. That is of course much better. will adjust it accordingly.

eerimoq: I hope it works as I didn't look too closely at it. Too much in Moblin, so hard to remember everything =)

Moblin/Various/Model.swift
```diff
@@ -1204,13 +1205,14 @@ final class Model: NSObject, ObservableObject {
...
+            guard let stream = self.getRtmpStream(streamKey: streamKey) else {
                 return
             }
-            self.media.removeRtmpCamera(cameraId: cameraId)
+            self.media.removeRtmpCamera(cameraId: stream.id)
             if self.database.debug!.enableRtmpAudio! {
-                self.media.removeRtmpAudio(cameraId: cameraId)
+                self.media.removeRtmpAudio(cameraId: stream.id)
             }
+            self.selectMic(mic: self.previousMic)
```
eerimoq, line 1215: What happens if the previous mic also was an RTMP mic, and that mic can no longer be used as that RTMP stream was disconnected some time ago?

eerimoq: Or if it was an bluetooth mic. The same logic can be applied to those I think.

eerimoq: It might be easiest to just start using a builtin mic if an external or RTMP mic is disconnected.

ErAzOr2k: I suggest I switch to the microphone, which is saved in the settings. that would be front, bottom or back

eerimoq: sounds good. didn't remember that exists =)

ErAzOr2k: or top of course :)

Moblin/Various/Model.swift
```diff
@@ -1204,13 +1204,14 @@ final class Model: NSObject, ObservableObject {
...
+            guard let stream = self.getRtmpStream(streamKey: streamKey) else {
                 return
             }
-            self.media.removeRtmpCamera(cameraId: cameraId)
+            self.media.removeRtmpCamera(cameraId: stream.id)
             if self.database.debug!.enableRtmpAudio! {
-                self.media.removeRtmpAudio(cameraId: cameraId)
+                self.media.removeRtmpAudio(cameraId: stream.id)
             }
+            self.setMicFromSettings()
```
eerimoq, line 1214: Maybe move this into the if statement above? The case I'm thinking about is is the user want to use an external bluetooth mic with RTMP video. If that case the external mic should continue to be used after the RTMP stream stops.

ErAzOr2k: OK, I changed it. had to add an additional check

PR #32, opened 2024-05-23
Add RTMP SampleBuffer Processor/Overhaul RTMP

ErAzOr2k:
This is a complete overhaul of the RTMP implementation.
Incoming sample buffers are now buffered for a certain time.

Benefits:
- Removed the hack that took the internal camera timestamps. That only led to problems anyway.
- Timestamps are now generated by fps for video
- Non-constant incoming sample buffers are now processed at the expected timing. fixes problem with DJI Pocket 3. at least in my tests
- RTMP video now runs smoothly and without stuttering
- RTMP Audio/Video now in Sync
- No need to set 0fps for b-frames

Tests and feedback is welcome.

### DIFF
--- a/Moblin/HaishinKit/AVCapture/MoblinCaptureDevice.swift
+++ b/Moblin/HaishinKit/AVCapture/MoblinCaptureDevice.swift
@@ -1,0 +1,8 @@
+//
+//  MoblinCaptureDevice.swift
+//  Moblin
+//
+//  Created by Markus Wölfer on 30.05.24.
+//
+
+import Foundation
--- a/Moblin/HaishinKit/AVCapture/MoblinCaptureDevice.swift
+++ b/Moblin/HaishinKit/AVCapture/MoblinCaptureDevice.swift
@@ -1,8 +1,0 @@
-//
-//  MoblinCaptureDevice.swift
-//  Moblin
-//
-//  Created by Markus Wölfer on 30.05.24.
-//
-
-import Foundation
--- a/Moblin/HaishinKit/Extension/Double+Extension.swift
+++ b/Moblin/HaishinKit/Extension/Double+Extension.swift
@@ -1,8 +1,0 @@
-import Foundation
-
-extension Double {
-    func nearestCommonFrameRate() -> Double {
-        let commonFrameRates = [25.0, 30.0, 50.0, 60.0]
-        return commonFrameRates.min(by: { abs($0 - self) < abs($1 - self) })!
-    }
-}
--- a/Moblin/HaishinKit/Extension/Double+Extension.swift
+++ b/Moblin/HaishinKit/Extension/Double+Extension.swift
@@ -2,7 +2,7 @@ import Foundation
 
 extension Double {
     func nearestCommonFrameRate() -> Double {
-        let commonFrameRates = [24.0, 25.0, 29.0, 30.0, 50.0, 60.0]
+        let commonFrameRates = [25.0, 30.0, 50.0, 60.0]
         return commonFrameRates.min(by: { abs($0 - self) < abs($1 - self) })!
     }
 }
--- a/Moblin/HaishinKit/Extension/Double+Extension.swift
+++ b/Moblin/HaishinKit/Extension/Double+Extension.swift
@@ -1,0 +1,8 @@
+import Foundation
+
+extension Double {
+    func nearestCommonFrameRate() -> Double {
+        let commonFrameRates = [24.0, 25.0, 29.0, 30.0, 50.0, 60.0]
+        return commonFrameRates.min(by: { abs($0 - self) < abs($1 - self) })!
+    }
+}
--- a/Moblin/HaishinKit/Media/AudioUnit.swift
+++ b/Moblin/HaishinKit/Media/AudioUnit.swift
@@ -37,9 +37,15 @@ private class ReplaceAudio {
         case buffering
         case outputting
     }
+
     private var maxQueueSize: Int {
-        return Int(latency * frameRate * 2)
+        return Int(latency * frameRate + 1)
     }
+
+    private var minQueueSize: Int {
+        return Int(latency * frameRate - 1)
+    }
+
     weak var delegate: ReplaceAudioSampleBufferDelegate?
 
     init(cameraId: UUID, latency: Double) {
@@ -53,19 +59,16 @@ private class ReplaceAudio {
             startTime = currentTime
         }
         sampleBufferQueue.append(sampleBuffer)
-        if state == .outputting, sampleBufferQueue.count > maxQueueSize {
-            logger.info("Too many ReplaceAudios buffered. Dropping oldest ReplaceAudio.")
-            sampleBufferQueue.removeFirst()
-        }
-        logger.info("ReplaceAudio Queue Count: \(sampleBufferQueue.count)")
+        // logger.info("ReplaceAudio Queue Count: \(sampleBufferQueue.count)")
+
         switch state {
         case .initializing:
             // logger.info("ReplaceVideo initializing.")
             if currentTime - startTime! >= initializationDuration {
                 initialize(sampleBuffer: sampleBuffer)
-                sampleBufferQueue.removeAll()
                 startTime = nil
                 state = .buffering
+                return
             }
         case .buffering:
             // logger.info("ReplaceAudio buffering.")
@@ -74,8 +77,8 @@ private class ReplaceAudio {
                 startOutput()
             }
         case .outputting:
-            // logger.info("Starting ReplaceAudio outputting.")
-            break
+            // logger.info("ReplaceAudio outputting.")
+            balanceQueue()
         }
     }
 
@@ -84,6 +87,7 @@ private class ReplaceAudio {
         let sampleRate = CMSampleBufferGetFormatDescription(sampleBuffer)?.streamBasicDescription?.pointee
             .mSampleRate
         frameRate = sampleRate! / frameLength
+        sampleBufferQueue.removeAll()
     }
 
     private func startOutput() {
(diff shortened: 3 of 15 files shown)


Moblin/RtmpServer/RtmpServerChunkStream.swift
```diff
@@ -471,40 +445,31 @@ class RtmpServerChunkStream {
...
-            duration -= Int64(videoTimestamp)
-            videoTimestamp = Double(messageTimestamp) - videoTimestampZero
-        } else {
-            videoTimestamp += Double(messageTimestamp)
-        }
-        var presentationTimeStamp: Int64
-        var decodeTimeStamp: Int64
-        if client.fps == 0 {
-            presentationTimeStamp = Int64(videoTimestamp) + Int64(compositionTime)
-            decodeTimeStamp = Int64(videoTimestamp)
```
eerimoq, line 499: Not handling PTS and DTS separetly will probably break video streams with B-frames. That is probably fine, as I'm sure very few, if any, of the RTMP publishers that feeds data to Mobli uses B-frames.

Moblin/HaishinKit/Media/VideoUnit.swift
```diff
@@ -257,6 +206,45 @@ final class VideoUnit: NSObject {
         _ = appendSampleBuffer(sampleBuffer, isFirstAfterAttach: false, applyBlur: ioVideoBlurSceneSwitch)
     }
 
+    private func startBlackImageTimer() {
```
eerimoq, line 209: If possible, to slightly reduce power usage, I suggest using a single timer instead of two (gap and black). Or is there some case we need two?

ErAzOr2k:
Unfortunately, DJI Action 4 and Pocket 3 behave completely differently.
While Action 4 runs perfectly in sync, the timestamps with the Pocket 3 drift apart. 
I'm currently running out of ideas on how to solve this :(

This is what Pocket 3 it looks like at the beginning:
23:11:03,500 Int. Video TimeStamp: 3701.372387333
23:11:03,535 Int. Video TimeStamp: 3701.405736958
23:11:03,569 Int. Video TimeStamp: 3701.439086458
23:11:03,601 Int. Video TimeStamp: 3701.472449125
23:11:03,635 Int. Video TimeStamp: 3701.505798625

few minutes later:
23:14:21,557 RTMP Video TimeStamp: 3893.161032392
23:14:21,584 Int. Video TimeStamp: 3899.466623875
23:14:21,602 RTMP Video TimeStamp: 3893.194365725
23:14:21,615 Int. Video TimeStamp: 3899.499958791
23:14:21,643 RTMP Video TimeStamp: 3893.227699058
23:14:21,648 Int. Video TimeStamp: 3899.533293625

And this with Action 4:
23:22:04,920 Int. Video TimeStamp: 139.21645325
23:22:04,928 RTMP Video TimeStamp: 139.314406901
23:22:04,948 RTMP Video TimeStamp: 139.347740234
23:22:04,957 Int. Video TimeStamp: 139.249787916
23:22:04,965 RTMP Video TimeStamp: 139.381073567

23:25:08,819 RTMP Video TimeStamp: 323.014405064
23:25:08,833 Int. Video TimeStamp: 323.124154583
23:25:08,840 RTMP Video TimeStamp: 323.047738397
23:25:08,863 RTMP Video TimeStamp: 323.08107173
23:25:08,864 Int. Video TimeStamp: 323.157489416
23:25:08,892 RTMP Video TimeStamp: 323.114405063

Of course I tested with 30fps on both devices...

Do you have other devices to test with?

eerimoq:
Rumors says RTMP in DJI products is bugged. I noticed it myself when testing with my DJI drone. The Pocket 3 probably outputs frames at a slightly different frame rate, or drops frames occasionally. So incrementing the PTS with `1 / fps` for every frame will likely not work for that camera.

It might be a good idea to start a periodic timer with `1 / fps` as timeout, and use that to pace it all (and discard/duplicate frames if needed). I mean, if I am not mistaken received frames are processed immediately in this PR, which means the video preview on phone screen will be glitchy if the RTMP connection has hickups. A smallish jitter buffer is probably a good idea to mitigate such problems.

eerimoq:
Another advantage of adding a timer is that the stream from Moblin might be configured for 60 FPS, but the RTMP server cam only inputs 30 FPS. So one have to duplicate frames to get to 60 FPS. This is only really needed if switching from RTMP cam to builtin cam during a stream.

ErAzOr2k:
you are a genius. That seems to be the right approach. because now it's much better. 
It's not quite perfect yet, but I think it can still be optimized. 
And yes, you're right, the preview will also be much smoother

eerimoq:
Amazing! I really appreciate you helping out making Moblin better!

Moblin/RtmpServer/RtmpServerChunkStream.swift
tt2468: I think b frames would be handy to have, especially given the quality benefit. Seems worthwhile to have, and separating PTS/DTS isn't really that complicated to implement.

Moblin/HaishinKit/Media/VideoUnit.swift
```diff
@@ -37,68 +37,83 @@ private struct FaceDetectionsCompletion {
...
-        else {
-            return nil
+    func startOutput() {
+        let frameInterval = 1 / frameRate
+        outputTimer = DispatchSource.makeTimerSource(queue: lockQueue)
+        outputTimer!.schedule(deadline: .now(), repeating: frameInterval)
+        outputTimer!.setEventHandler { [weak self] in
+            self?.outputSampleBuffer()
+        }
+        outputTimer!.activate()
```
eerimoq, line 85: This outputs frames at the configured RTMP server stream FPS, which makes sense. The problem is that this output rate makes the live stream (out of Moblin) the same FPS, not the FPS configured in Settings -> Streams -> My stream -> Video -> FPS. The face effect expects frames to arrive at live stream FPS, so its timing will be incorrect. I haven't made up my mind if it is the face effect implementation that should be changed to fix this. I do think so, I think =)

Long term multi camera may be a thing, and that of course means merging multiple video feeds, with potentially different FPS:s, into a single feed with one FPS. But that's too far into the future to take into account now.

eerimoq (commented):
Haven't tested the PR yet. Just wrote some general comments.

eerimoq:
Does the user have to configure the correct RTMP server stream FPS now? Settings it to 0 will not work? Just asking.

ErAzOr2k:
> Does the user have to configure the correct RTMP server stream FPS now? Settings it to 0 will not work? Just asking.
> 
> If the configured RTMP server stream FPS is set lower than the real FPS produced by the RTMP publisher, Moblin will run out of memory as `sampleBufferQueue` in videounit will grow.

yes, at this state of the PRs it is still necessary. This controls the processing speed of the queue. But since we can now do without specifying the fps when calculating the timestamps, I think I will find a solution here too. so that fps no longer needs to be set manually in the future

eerimoq:
The timestamps received over RTMP from my DJI drone were incorrect last time I tested and to get any sense of timing the excepted FPS has to be configured somehow. Or possibly estimated over time.

Moblin/HaishinKit/Media/VideoUnit.swift
eerimoq: As a builtin camera is used for output FPS it will follow that. The RTMP frames will be duplicated of dropped if needed.

Moblin/HaishinKit/AVCapture/MoblinCaptureDevice.swift
```diff
@@ -0,0 +1,8 @@
+//
+//  MoblinCaptureDevice.swift
+//  Moblin
+//
+//  Created by Markus Wölfer on 30.05.24.
+//
+
+import Foundation
```
eerimoq, line 8: Why this empty file?

Moblin/HaishinKit/Extension/Double+Extension.swift
```diff
@@ -0,0 +1,8 @@
+import Foundation
+
+extension Double {
+    func nearestCommonFrameRate() -> Double {
```
eerimoq, line 4: In my opinion it doesn't make sense to make this an extension of Double. The functionality is not related to the type, it just happens to use double.

Moblin/HaishinKit/Media/AudioUnit.swift
```diff
@@ -19,26 +19,120 @@ func makeChannelMap(
...
+            startTime = currentTime
+        }
+        sampleBufferQueue.append(sampleBuffer)
+        // logger.info("ReplaceAudio Queue Count: \(sampleBufferQueue.count)")
+
+        switch state {
+        case .initializing:
+            // logger.info("ReplaceVideo initializing.")
+            if currentTime - startTime! >= initializationDuration {
+                initialize(sampleBuffer: sampleBuffer)
```
eerimoq, line 68: What is TCP has issues and only two frames have been received after 1 second? A bit too fragile I think.

Moblin/HaishinKit/Media/AudioUnit.swift
```diff
@@ -19,26 +19,120 @@ func makeChannelMap(
...
+    private var initializationDuration: Double = 1
+    private var outputTimer: DispatchSourceTimer?
+    private enum State {
+        case initializing
+        case buffering
+        case outputting
+    }
+
+    private var maxQueueSize: Int {
+        return Int(latency * frameRate + 1)
```
eerimoq, line 42: Is it enough to only allow a single frame more than expected to to buffered? I mean, there can be bursts of packets and frames from the RTMP publisher, which could mean we want to buffer more than 1 frame more than the target buffer size.

Moblin/HaishinKit/Media/AudioUnit.swift
ErAzOr2k: So you would make it more dependent on the count of frames?

eerimoq: Sounds better, yes. The alternative is to set the timer interval to match the stream output fps.

Moblin/HaishinKit/AVCapture/MoblinCaptureDevice.swift
ErAzOr2k: pushed by mistake. File is now deleted

Moblin/HaishinKit/Extension/Double+Extension.swift
ErAzOr2k: done

PR #38, opened 2024-06-11
RTMP Rework

ErAzOr2k:


Moblin/Various/Settings.swift
```diff
@@ -1125,7 +1125,10 @@ class SettingsRtmpServerStream: Codable, Identifiable {
     var name: String = "My stream"
     var streamKey: String = ""
     var latency: Int32? = defaultRtmpLatency
+    var manualFps: Bool? = false
     var fps: Double? = 0
+    var autoSelectRtmpScene: Bool? = false
+    var autoSelectRtmpMic: Bool? = false
```
eerimoq, line 1131: I prefer not adding the auto selection settings now. Maybe ok with mic, but not scene as there is no automatic scene selection anywhere else in Moblin. Can maybe add leter, but let's not do it now.

Moblin/HaishinKit/Codec/VideoCodec.swift
```diff
@@ -133,6 +135,10 @@ class VideoCodec {
         ] status, _, imageBuffer, presentationTimeStamp, duration in
             guard let imageBuffer, status == noErr else {
                 logger.info("video: Failed to decode frame status \(status)")
+                guard let lastSampleBuffer else {
+                    return
+                }
+                delegate?.videoCodecOutputSampleBuffer(self, lastSampleBuffer)
```
eerimoq, line 141: Is it really needed to output the last sample buffer? I mean, doesn't the sample buffer queue in the video unit already output the last buffer?

Moblin/Various/Settings.swift
ErAzOr2k: I can't quite understand that. Because from the user's perspective, if I connect an RTMP publisher, I will also want to stream through it. Every click that can be saved contributes to usability. especially when the decision is made to the user. but you are the boss :) I have now taken it out.

Moblin/Various/Settings.swift
```diff
@@ -2578,5 +2584,21 @@ final class Settings {
             stream.backgroundStreaming = false
             store()
         }
+//        for stream in realDatabase.rtmpServer!.streams where stream.buggedPublisher == nil {
+//            stream.buggedPublisher = false
+//            store()
+//        }
```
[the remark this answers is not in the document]
ErAzOr2k: done

Moblin/Various/Settings.swift
```diff
@@ -2578,5 +2584,21 @@ final class Settings {
...
+        for stream in realDatabase.rtmpServer!.streams where stream.manualFps == nil {
+            stream.manualFps = false
+            store()
+        }
```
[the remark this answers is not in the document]
ErAzOr2k: done

Moblin/HaishinKit/Codec/VideoCodec.swift
ErAzOr2k: That's right, with the current implementation this is no longer necessary. took it out

Moblin/Various/Settings.swift
eerimoq: thanks, I'm sure we will revisit it later

PR #41, opened 2024-06-14
Belabox inspired adaptive bitrate algorithm

jui:


eerimoq:
Looks awesome. Let's add a comment in the beginning of the bela-file with links and such to the original algorithm.

mikeful:
Testing on my phone felt relatively stable. Bitrate increases are slower than current Moblin algorithm on fast preset.

eerimoq:
Ok! Will give it a try myself tomorrow as I'm at DH today. 

I'm sure the bela algorithm works well, so I think we can proceed adding it to the adaptive bitrate algorithm drop down. Let's keep the current algorithms as well to make it easy to compare and keep the current Fast IRL algorithm as default for now.

PR #43, opened 2024-06-15
jitter optimization

ErAzOr2k:


Moblin/HaishinKit/Media/VideoUnit.swift
```diff
@@ -41,20 +41,31 @@ private struct FaceDetectionsCompletion {
...
     }
 
-    func updateSampleBuffer(_ realPresentationTimeStamp: Double) {
+    func updateSampleBuffer() {
+        realPresentationTimeStamp = CMTimeAdd(
+            realPresentationTimeStamp,
+                   CMTime(
+                       value: CMTimeValue(Double(1)),
+                       timescale: CMTimeScale(frameRate)
+                   ))
```
eerimoq, line 67: I think it is better tot calculate the PTS from a frame counter for less drift.

ErAzOr2k: builtin camera should now work after background/foreground with latest commit

PR #47, opened 2024-07-03
Disable internal camera on rtmp scene

ErAzOr2k:


Moblin/Various/Media.swift
```diff
@@ -739,8 +739,8 @@ final class Media: NSObject {
         })
     }
 
-    func attachReplaceCamera(cameraId: UUID, device: AVCaptureDevice?) {
-        netStream.attachCamera(device, replaceVideoCameraId: cameraId)
```
eerimoq, line 743: It is tempting to add a new attach function `NetStream.attachReplaceCamera()`, just as you did, but let's do the minimum, which is to simply pass `device` as `nil` to `NetStream.attachCamera()`. That will detach any builtin camera, and at the same time attach the "replace camera". No changes are needed in `NetStream.swift`, `Mixer.swift`, `VideoUnit.swift` if I'm not mistaken.

ErAzOr2k: ok changed it to the minimal solution :)

PR #49, opened 2024-07-04
Remote Control Info and Preview on Watch

ErAzOr2k:
Display data and preview from the remote control streamer instead of from the device connected to the watch. You have to enable this in the watch settings.

### DIFF
--- a/Common/Various/CommonUtils.swift
+++ b/Common/Various/CommonUtils.swift
@@ -507,34 +507,8 @@ func formatAudioLevelDb(level: Float) -> String {
     return String(localized: "\(Int(level)) dB,")
 }
 
-func formatAudioLevel(level: Float) -> String {
-    if level.isNaN {
-        return "Muted,"
-    } else if level == .infinity {
-        return "Unknown,"
-    } else {
-        return formatAudioLevelDb(level: level)
-    }
-}
-
 func formatAudioLevelChannels(channels: Int) -> String {
     return String(localized: " \(channels) ch")
-}
-
-func parseAudioLevel(_ string: String?) -> Float {
-    guard let string = string else {
-        return Float.infinity
-    }
-    if string.hasPrefix("Muted,") {
-        return Float.nan
-    } else if string.hasPrefix("Unknown,") {
-        return Float.infinity
-    } else if let range = string.range(of: " dB,") {
-        let levelString = String(string[string.startIndex ..< range.lowerBound])
-        return Float(levelString) ?? Float.nan
-    } else {
-        return Float.nan
-    }
 }
 
 let noValue = ""
--- a/Common/Various/CommonUtils.swift
+++ b/Common/Various/CommonUtils.swift
@@ -521,6 +521,22 @@ func formatAudioLevelChannels(channels: Int) -> String {
     return String(localized: " \(channels) ch")
 }
 
+func parseAudioLevel(_ string: String?) -> Float {
+    guard let string = string else {
+        return Float.infinity
+    }
+    if string.hasPrefix("Muted,") {
+        return Float.nan
+    } else if string.hasPrefix("Unknown,") {
+        return Float.infinity
+    } else if let range = string.range(of: " dB,") {
+        let levelString = String(string[string.startIndex ..< range.lowerBound])
+        return Float(levelString) ?? Float.nan
+    } else {
+        return Float.nan
+    }
+}
+
 let noValue = ""
 
 func urlImage(interfaceType: NWInterface.InterfaceType) -> String {
--- a/Moblin Watch/Various/Model.swift
+++ b/Moblin Watch/Various/Model.swift
@@ -249,7 +249,9 @@ class Model: NSObject, ObservableObject {
     }
 
     private func handleThermalState(_ data: Any) throws {
-        guard let thermalState = data as? ProcessInfo.ThermalState else {
+        guard let value = data as? Int,
+              let thermalState = ProcessInfo.ThermalState(rawValue: value)
+        else {
             logger.info("Invalid thermal state message")
             return
         }
--- a/Moblin Watch/Various/Model.swift
+++ b/Moblin Watch/Various/Model.swift
@@ -246,11 +246,6 @@ class Model: NSObject, ObservableObject {
         if self.settings.show == nil {
             self.settings.show = .init()
         }
-        if self.settings.remoteControl == true {
-            previewTimeout = Duration.seconds(5)
-        } else {
-            previewTimeout = Duration.seconds(10)
-        }
     }
 
     private func handleThermalState(_ data: Any) throws {
--- a/Moblin Watch/Various/Model.swift
+++ b/Moblin Watch/Various/Model.swift
@@ -3,7 +3,7 @@ import Foundation
 import SwiftUI
 import WatchConnectivity
 
-private let previewTimeout = Duration.seconds(5)
+private var previewTimeout = Duration.seconds(5)
 
 struct ChatPostSegment: Identifiable {
     var id = UUID()
@@ -245,6 +245,11 @@ class Model: NSObject, ObservableObject {
         }
         if self.settings.show == nil {
             self.settings.show = .init()
+        }
+        if self.settings.remoteControl == true {
+            previewTimeout = Duration.seconds(5)
+        } else {
+            previewTimeout = Duration.seconds(10)
         }
     }
 
--- a/Moblin Watch/Various/Model.swift
+++ b/Moblin Watch/Various/Model.swift
@@ -3,7 +3,7 @@ import Foundation
 import SwiftUI
 import WatchConnectivity
 
-private var previewTimeout = Duration.seconds(5)
+private let previewTimeout = Duration.seconds(5)
 
 struct ChatPostSegment: Identifiable {
     var id = UUID()
--- a/Moblin/Various/Model.swift
+++ b/Moblin/Various/Model.swift
@@ -4970,42 +4970,42 @@ extension Model {
     }
 
     func updateRemoteControlAssistantStatus() {
-        guard remoteControlAssistant?.isConnected() == true else {
+        guard showingRemoteControl, remoteControlAssistant?.isConnected() == true else {
             return
         }
         remoteControlAssistant?.getStatus { general, topLeft, topRight in
             self.remoteControlGeneral = general
             self.remoteControlTopLeft = topLeft
             self.remoteControlTopRight = topRight
+            if self.isRemoteControlAssistantConnected() {
+                var thermalState: ProcessInfo.ThermalState
+                switch self.remoteControlGeneral?.flame {
+                case .white:
+                    thermalState = .fair
+                case .yellow:
+                    thermalState = .serious
+                case .red:
+                    thermalState = .critical
+                case .none:
+                    thermalState = .critical
+                }
+                self.sendThermalStateToWatch(thermalState: thermalState)
+                if let recordingMessage = topRight.recording?.message {
+                    self.sendRecordingLengthToWatch(recordingLength: recordingMessage)
+                }
+                if let bitrateMessage = topRight.bitrate?.message {
+                    self.sendSpeedAndTotalToWatch(speedAndTotal: bitrateMessage)
+                }
+                if let audioLevel = topRight.audioLevel {
+                    self.sendAudioLevelToWatch(audioLevel: audioLevel)
+                    self.sendIsMutedToWatch(isMuteOn: audioLevel.isNaN)
+                }
+                self.sendIsLiveToWatch(isLive: topRight.bitrate?.message != nil)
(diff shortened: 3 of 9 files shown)


Common/Various/CommonUtils.swift
```diff
@@ -521,6 +521,22 @@ func formatAudioLevelChannels(channels: Int) -> String {
...
+    } else if string.hasPrefix("Unknown,") {
+        return Float.infinity
+    } else if let range = string.range(of: " dB,") {
+        let levelString = String(string[string.startIndex ..< range.lowerBound])
+        return Float(levelString) ?? Float.nan
+    } else {
+        return Float.nan
+    }
+}
+
```
eerimoq, line 539: We should add the numeric value to the remote control protocol to simplify this and make it work with bars.

Moblin Watch/Various/Model.swift
```diff
@@ -246,6 +246,11 @@ class Model: NSObject, ObservableObject {
         if self.settings.show == nil {
             self.settings.show = .init()
         }
+        if self.settings.remoteControl == true {
+            previewTimeout = Duration.seconds(5)
+        } else {
+            previewTimeout = Duration.seconds(10)
+        }
```
eerimoq, line 253: why is this needed?

Moblin/Various/Model.swift
```diff
@@ -4966,13 +4980,50 @@ extension Model {
...
+            }
+            if let bitrateMessage = remoteControlTopRight?.bitrate?.message {
+                sendSpeedAndTotalToWatch(speedAndTotal: bitrateMessage)
+            }
+            sendThermalStateToWatch(thermalState: thermalState)
+            sendAudioLevelToWatch(audioLevel: audioLevel)
+            sendIsLiveToWatch(isLive: remoteControlTopRight?.bitrate?.message != nil)
+            sendIsRecordingToWatch(isRecording: remoteControlTopRight?.recording?.message != nil)
+            sendIsMutedToWatch(isMuteOn: audioLevel.isNaN)
+        }
```
eerimoq, line 5007: Should this be sent when the remote control assistant receives updates, not send the previous update?

Moblin/Various/Model.swift
```diff
@@ -5076,32 +5130,32 @@ extension Model {
         )
     }
 
-    private func sendSpeedAndTotalToWatch() {
+    private func sendSpeedAndTotalToWatch(speedAndTotal: Any) {
```
eerimoq, line 5133: Use `Any` as little as possible for lower risk of mistakes.

Moblin/Various/Model.swift
```diff
@@ -5274,11 +5330,16 @@ extension Model: WCSessionDelegate {
...
+        if database.watch!.remoteControl! {
+            remoteControlAssistantSetStream(on: value)
+            sendIsLiveToWatch(isLive: value)
+        } else {
+            DispatchQueue.main.async {
+                if value {
+                    self.startStream()
+                } else {
+                    self.stopStream()
+                }
```
eerimoq, line 5342: The whole if-else should be in main async block, right?

Moblin/Various/Model.swift
```diff
@@ -5287,11 +5348,16 @@ extension Model: WCSessionDelegate {
...
+        if database.watch!.remoteControl! {
+            remoteControlAssistantSetRecord(on: value)
+            sendIsRecordingToWatch(isRecording: value)
+        } else {
+            DispatchQueue.main.async {
+                if value {
+                    self.startRecording()
+                } else {
+                    self.stopRecording()
+                }
```
eerimoq, line 5360: Same comment as above.

Moblin/Various/Model.swift
```diff
@@ -5300,8 +5366,13 @@ extension Model: WCSessionDelegate {
...
         }
-        DispatchQueue.main.async {
-            self.setIsMuted(value: value)
+        if database.watch!.remoteControl! {
+            remoteControlAssistantSetMute(on: value)
+            sendIsMutedToWatch(isMuteOn: value)
+        } else {
+            DispatchQueue.main.async {
+                self.setIsMuted(value: value)
+            }
```
eerimoq, line 5375: Same comment as above.

Moblin/View/Settings/Watch/WatchSettingsView.swift
```diff
@@ -15,6 +15,19 @@ struct WatchSettingsView: View {
...
+                    model.database.watch!.remoteControl!
+                }, set: { value in
+                    model.database.watch!.remoteControl = value
+                    model.store()
+                }))
+            } footer: {
+                Text("""
+                Display data and preview from the remote control streamer \
+                instead of from the device connected to the watch.
+                """)
```
eerimoq, line 29: I think this should say something like "Use the watch as remote control assistant instead of ...", or at least mention the assistant, as it is the assistant functionality we expose in the watch.

Common/Various/CommonUtils.swift
ErAzOr2k: Still in progress

Moblin Watch/Various/Model.swift
ErAzOr2k: The interval for the clock is 5 seconds. For the remote assistant, it's 10 seconds. I have now also changed the interval for the remote assistant to 5 seconds. I hope that's okay.

Moblin/Various/Model.swift
ErAzOr2k: I don't quite understand. The listed conditions are necessary to set the toggles.

Moblin/Various/Model.swift
ErAzOr2k: done

Moblin/Various/Model.swift
ErAzOr2k: done

Moblin/Various/Model.swift
ErAzOr2k: done

Moblin/Various/Model.swift
ErAzOr2k: done

Moblin/View/Settings/Watch/WatchSettingsView.swift
ErAzOr2k: I have written a new text

Common/Various/CommonUtils.swift
ErAzOr2k: Done. Although I'm not entirely satisfied because the alignment doesn't quite fit (compared to all the other elements).

What do you think? should we recreate a new audiolevenview for remotecontrol?

eerimoq: We want the UI to be pretty. So I guess yes =)

eerimoq: Or possibly send both what we have today and the numeric level.

ErAzOr2k: ok much cleaner now :)

Moblin/Various/Model.swift
eerimoq: `updateRemoteControlAssistantStatus()` requests an update from the streamer. It's the completion blocks for `getStatus()` and `getSettings()` that receives the responses, and it is in those completion blocks we should update the watch.

Moblin/Various/Model.swift
```diff
@@ -4955,7 +4970,7 @@ extension Model {
     }
 
     func updateRemoteControlAssistantStatus() {
-        guard showingRemoteControl, remoteControlAssistant?.isConnected() == true else {
```
eerimoq, line 4958: I would prefer the remote control assistant to only fetch information from the remote control streamer when the assistant is actually in use, to minimize the network traffic from the streamer.

ErAzOr2k: ok reverted it

Moblin/Various/Model.swift
ErAzOr2k: hope it's okay now :)

Moblin/Various/Model.swift
```diff
@@ -5288,10 +5351,15 @@ extension Model: WCSessionDelegate {
             return
         }
         DispatchQueue.main.async {
-            if value {
-                self.startRecording()
+            if self.isRemoteControlAssistantConnected() {
+                self.remoteControlAssistantSetRecord(on: value)
+                self.sendIsRecordingToWatch(isRecording: value)
```
eerimoq, line 5356: I can't remember the details, but if possible, we should not update the clock record state here, but instead when the streamer phone has responded and says it is recording.

Moblin/Various/Model.swift
```diff
@@ -4962,6 +4977,32 @@ extension Model {
             self.remoteControlGeneral = general
             self.remoteControlTopLeft = topLeft
             self.remoteControlTopRight = topRight
+            if self.isRemoteControlAssistantConnected() {
```
eerimoq, line 4980: This check can be removed as the same check is done in the beginning of the function.

Moblin/Various/Model.swift
ErAzOr2k: betrifft das auch handleSetIsMuted und handleSetIsLive?

Moblin/Various/Model.swift
ErAzOr2k: done

Moblin/Various/Model.swift
eerimoq: I think so, yes. But as I said, I can't remember is it's easy to do the the "correct" way. So don't put too much time into it, but give it a try.

ErAzOr2k: complete. I also fixed a serious bug with audio level

Moblin/Various/Model.swift
```diff
@@ -4678,9 +4694,8 @@ extension Model: RemoteControlStreamerDelegate {
             }
             var topRight = RemoteControlStatusTopRight()
             if self.isShowingStatusAudioLevel() {
-                let level = formatAudioLevel(level: self.audioLevel) +
-                    formatAudioLevelChannels(channels: self.numberOfAudioChannels)
-                topRight.audioLevel = RemoteControlStatusItem(message: level)
+                topRight.audioLevel = self.audioLevel.isNaN ? nil : self.audioLevel
```
eerimoq, line 4697: What is the reason for assigning `self.audioLevel` conditionally?

Moblin/Various/Model.swift
```diff
@@ -4955,13 +4970,37 @@ extension Model {
...
+            }
+            self.sendThermalStateToWatch(thermalState: thermalState)
+            if let recordingMessage = topRight.recording?.message {
+                self.sendRecordingLengthToWatch(recordingLength: recordingMessage)
+            }
+            self.sendIsRecordingToWatch(isRecording: topRight.recording?.message != nil)
+            if let bitrateMessage = topRight.bitrate?.message {
+                self.sendSpeedAndTotalToWatch(speedAndTotal: bitrateMessage)
+            }
+            self.sendIsLiveToWatch(isLive: topRight.bitrate?.message != nil)
```
eerimoq, line 4999: Why it `bitrate` used to check if live? Should we add the live state to the remote control protocol?

Moblin/Various/Model.swift
```diff
@@ -4955,13 +4970,37 @@ extension Model {
...
+            if let audioLevel = topRight.audioLevel {
+                self.sendAudioLevelToWatch(audioLevel: audioLevel)
+            }
+            self.sendIsMutedToWatch(isMuteOn: topRight.audioLevel == nil)
```
eerimoq, line 5003: Isn't the audio level NaN when muted?

Moblin/Various/Model.swift
ErAzOr2k: The reason is, when audio is muted (audiolevel = .nan) updateRemoteControlAssistantStatus won't get called. The reason seems to be that nan is noch decodeable/encodeable. took me a while to figure it out.

Moblin/Various/Model.swift
eerimoq: Might even be possible yo remote isMuted fro mthe watch protocol now that we have audioLevel. It replaces it I think.

ErAzOr2k: doesn't work with .nan

Moblin/Various/Model.swift
eerimoq: What about using an enum to represent the audio level? I have not tested it, but should work. And makes it more clear than using `nil`. Does infinity (as in unknown) work when encoding a float using JSON? Probably not?
``` swift
enum AudioLevel: Codable {
    case muted
    case unknown
    case value(Float)
}
```

ErAzOr2k: it works :)

Moblin/Various/Model.swift
ErAzOr2k: done

PR #54, opened 2024-07-15
better timestamp handling + FPS picker in rtmp server settings

ErAzOr2k:
- Modeled after FFmpeg and GStreamer, the dynamic timeBase is used to calculate Presentation Time Stamps (PTS) based on frames per second (fps)
- FPS picker in rtmp server settings
41F5AD68-8352-48D5-92F2-0A2FBC44CE88


ErAzOr2k:
I create a new PR tomorrow

PR #55, opened 2024-07-16
better timestamp handling + FPS picker in rtmp server settings

ErAzOr2k:
- Modeled after FFmpeg and GStreamer, the dynamic timeBase is used to calculate Presentation Time Stamps (PTS) based on frames per second (fps).
04:38:01,890 CapturePts: audio: 22336.562444833, video: 22336.55613725, delta: 0.006307583000307204 after 5h RTMP stream

- FPS picker in rtmp server settings
41F5AD68-8352-48D5-92F2-0A2FBC44CE88


Moblin/View/Settings/RtmpServer/RtmpServerStreamSettingsView.swift
```diff
@@ -136,12 +133,13 @@ struct RtmpServerStreamSettingsView: View {
                 }))
                 .disabled(model.rtmpServerEnabled())
                 if stream.manualFps! {
-                    TextEditNavigationView(
-                        title: String(localized: "FPS"),
-                        value: String(stream.fps!),
-                        onSubmit: submitFps,
-                        keyboardType: .numbersAndPunctuation
-                    )
+                    Picker(String(localized: "FPS"), selection: Binding(get: {
```
eerimoq, line 136: `String(localized: "FPS")` can be replaced with `"FPS"` here.

Moblin/View/Settings/RtmpServer/RtmpServerStreamSettingsView.swift
```diff
@@ -136,12 +133,13 @@ struct RtmpServerStreamSettingsView: View {
...
+                        String(stream.fps!)
+                    }, set: submitFps)) {
+                        ForEach(netStreamFps, id: \.self) {
+                            Text($0)
+                        }
+                    }
```
eerimoq, line 142: Floating point numbers are dangerous. You sure no weird rounding errors will occur?

Moblin/Various/Settings.swift
```diff
@@ -1319,13 +1319,15 @@ class SettingsDebug: Codable {
...
+let netStreamFps = ["25.0", "29.97", "30.0", "50.0", "59.94", "60.0"]
+
 class SettingsRtmpServerStream: Codable, Identifiable {
     var id: UUID = .init()
     var name: String = "My stream"
     var streamKey: String = ""
     var latency: Int32? = defaultRtmpLatency
     var manualFps: Bool? = false
-    var fps: Double? = 30
+    var fps: Double? = 30.0
```
eerimoq, line 1330: Let's keep fps as is, and add a new variable `selectedFps`, which is an enum. Just to make sure no rounding errors occurs.

eerimoq: We should also migrate and old setting to the new settings, so users that have configured an FPS will not lose it.

ErAzOr2k: done

Moblin/View/Settings/RtmpServer/RtmpServerStreamSettingsView.swift
ErAzOr2k: done

PR #58, opened 2024-07-24
Fix RTMP Audio crackle when in background

ErAzOr2k:
...and hopefully less power consumption, since the system time is not queried for each buffer but is calculated instead.


eerimoq:
Did you test switching to a scene with a builtin cam after streaming for a long time? I had problems previously that the timestamps mismatched and video freezed.

PR #64, opened 2024-09-10
fix: lower deployment target

SputNikPlop:
This should enable more devices and to let me at least work on this via TestFlight

### DIFF
--- a/Moblin.xcodeproj/project.pbxproj
+++ b/Moblin.xcodeproj/project.pbxproj
@@ -3233,7 +3233,7 @@
 				INFOPLIST_KEY_NSMicrophoneUsageDescription = Yes;
 				INFOPLIST_KEY_NSPhotoLibraryAddUsageDescription = Yes;
 				INFOPLIST_KEY_NSPhotoLibraryUsageDescription = Yes;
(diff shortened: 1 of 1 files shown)


eerimoq:
Which devices do support iOS 16.0 but not iOS 16.4?

SputNikPlop:
> Which devices do support iOS 16.0 but not iOS 16.4?

There is no particular device, it really comes down to software issue during updates for example https://discussions.apple.com/thread/254244622?sortBy=rank iOS 16.0 on a test device works, 16.0.3 on another doesn't work then it breaks later on another release. Main goal is lettings this work for accessibility reasons.

eerimoq:
image
Have to rebuild libsrt it looks like.

Baconing:
Seconded, I have an older iOS version because my phone is jailbroken, and it would be nice to keep my older version.

eerimoq:
Will likely stay at 16.4 for some time, and then bump to 17+. No time frame really

SputNikPlop:
> Seconded, I have an older iOS version because my phone is jailbroken, and it would be nice to keep my older version.

What version is your jailbroken phone?

PR #71, opened 2024-10-10
Simplified Chinese translation updated.

WanQiyang:
The further refinements in the Simplified Chinese translation.


eerimoq:
Thank you!

PR #103, opened 2025-03-04
Retain active scene on deletion of widget

lahariganti:


eerimoq:
Thanks!

What happens if you delete the currently active scene?

eerimoq:
Ahh widget deletion. Misread

lahariganti:
> Thanks!
> 
> What happens if you delete the currently active scene?

Have updated the PR description with screen recordings. Defaults to the next one in the list.

eerimoq:
After reordering the scenes, the UI says Front scene is selected, but the back camera is active.

PR #145, opened 2025-08-06
Cleaned up Toast for Kick

id3adeye:


eerimoq:
hmm maybe we should replace the current highlight image icon with "text/unicode icons" instead of having both as in this PR?

PR #146, opened 2025-08-08
Fix Chat Reply Emotes Display

id3adeye:
Problem: Chat reply previews showed [emote] text instead of actual emote images.

  Solution:
  - Updated ChatHighlight to use segment-based rendering like regular messages
  - Fixed Twitch reply emote processing bug
  - Now shows actual emote images in reply previews across all chat views

### DIFF
--- a/Moblin/Platforms/Twitch/TwitchChat.swift
+++ b/Moblin/Platforms/Twitch/TwitchChat.swift
@@ -596,7 +596,7 @@ final class TwitchChat {
         webSocket.send(string: "PONG \(message.parameters.joined(separator: " "))")
     }
 
-    private func createHighlight(message: ChatMessage, emotes: [ChatMessageEmote]) -> ChatHighlight? {
+    private func createHighlight(message: ChatMessage, emotes _: [ChatMessageEmote]) -> ChatHighlight? {
         if message.announcement {
             return ChatHighlight.makeAnnouncement()
         } else if message.firstMessage {
@@ -604,7 +604,7 @@ final class TwitchChat {
         } else if let sender = message.replySender, let text = message.replyText {
             return ChatHighlight.makeReply(
                 user: sender,
-                segments: createSegments(text: text, emotes: emotes, emotesManager: self.emotes, bits: nil)
+                segments: createSegments(text: text, emotes: [], emotesManager: emotes, bits: nil)
             )
         } else {
             return nil
--- a/Moblin/Platforms/Twitch/TwitchChat.swift
+++ b/Moblin/Platforms/Twitch/TwitchChat.swift
@@ -574,7 +574,7 @@ final class TwitchChat {
             isSubscriber: message.subscriber,
             isModerator: message.moderator,
             bits: message.bits,
-            highlight: createHighlight(message: message, emotes: emotes)
+            highlight: createHighlight(message: message)
         )
     }
 
@@ -596,7 +596,7 @@ final class TwitchChat {
         webSocket.send(string: "PONG \(message.parameters.joined(separator: " "))")
     }
 
-    private func createHighlight(message: ChatMessage, emotes _: [ChatMessageEmote]) -> ChatHighlight? {
+    private func createHighlight(message: ChatMessage) -> ChatHighlight? {
         if message.announcement {
             return ChatHighlight.makeAnnouncement()
         } else if message.firstMessage {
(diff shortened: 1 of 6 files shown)


Common/Localizable.xcstrings
```diff
@@ -37245,6 +37245,7 @@
       }
     },
     "Chat is empty." : {
+      "extractionState" : "stale",
```
eerimoq, line 37248: None of these should be stale. They are all used in the Watch app.

Moblin/Platforms/Twitch/TwitchChat.swift
```diff
@@ -596,15 +596,15 @@ final class TwitchChat {
         webSocket.send(string: "PONG \(message.parameters.joined(separator: " "))")
     }
 
-    private func createHighlight(message: ChatMessage, emotes: [ChatMessageEmote]) -> ChatHighlight? {
+    private func createHighlight(message: ChatMessage, emotes _: [ChatMessageEmote]) -> ChatHighlight? {
```
eerimoq, line 599: Remove the `emotes` parameter as it is not needed anymore.

Moblin/Various/ChatPost.swift
```diff
@@ -36,66 +36,102 @@ struct ChatHighlight {
...
     let barColor: Color
     let image: String
-    let title: String
+    let titleSegments: [ChatPostSegment]
+
+    var title: String {
+        titleSegments.compactMap { segment in
+            segment.text ?? (segment.url != nil ? "[emote]" : nil)
+        }.joined()
+    }
```
eerimoq, line 45: Do we really want to show `[emote]` on watch and on external display?

Moblin/Various/ChatPost.swift
```diff
@@ -36,66 +36,102 @@ struct ChatHighlight {
...
 
     static func makeReply(user: String, segments: [ChatPostSegment]) -> ChatHighlight {
-        var title = String(localized: "Replying to \(user):")
+        let prefixText = String(localized: "Replying to \(user): ")
+        var replySegments: [ChatPostSegment] = []
+        var id = 0
+
+        // Add prefix text segment
```
eerimoq, line 52: Remove blank lines within functions and comments that are not super important.

Moblin/Various/ChatPost.swift
```diff
@@ -36,66 +36,102 @@ struct ChatHighlight {
...
+                        let truncatedText = String(text.prefix(remainingLength - 3)) + "..."
+                        replySegments.append(ChatPostSegment(id: id, text: truncatedText))
+                    } else {
+                        replySegments.append(ChatPostSegment(id: id, text: "..."))
+                    }
+                    break
+                }
+                totalLength += text.count
+                replySegments.append(ChatPostSegment(id: id, text: segment.text, url: segment.url))
+            } else if segment.url != nil {
```
eerimoq, line 72: Hmm I think `segment.text` and `segment.url` can be present in the same segment. It's a bit odd, but it can happen, at least in some places. Not sure if it applies here, but maybe.

Moblin/Various/ChatPost.swift
```diff
@@ -36,66 +36,102 @@ struct ChatHighlight {
...
+                if totalLength + 7 > 65 { // "[emote]" length
```
eerimoq, line 73: Is `[emote]` ever part of a segment here? It's only added in the title string of what I can find.

Moblin/Various/ChatPost.swift
```diff
@@ -36,66 +36,102 @@ struct ChatHighlight {
...
+                    replySegments.append(ChatPostSegment(id: id, text: "..."))
+                    break
+                }
+                totalLength += 1 // Emote takes minimal space in counting
```
eerimoq, line 77: 1 space seems too little. maybe 3?

Moblin/View/ControlBar/QuickButton/QuickButtonChatView.swift
```diff
@@ -15,7 +19,24 @@ private struct HighlightMessageView: View {
         ) {
             Image(systemName: highlight.image)
             Text(" ")
-            Text(highlight.title)
+
```
eerimoq, line 22: Remove blank line.

Moblin/View/Stream/Overlay/StreamOverlayChatView.swift
```diff
@@ -34,9 +38,34 @@ private struct HighlightMessageView: View {
         ) {
             Image(systemName: highlight.image)
             Text(" ")
-            Text(highlight.title)
+
```
eerimoq, line 41: Remove blank line.

PR #158, opened 2025-09-03
channel ID display and rate limiting to the Kick settings:

id3adeye:
Channel ID Display: Shows the fetched channel ID under the channel name
   field, or "Channel not found" if invalid
Rate Limiting: 2.5-second cooldown on channel name changes to prevent
  API rate limiting


eerimoq:
I think we should start by fetching the channel info in one place in the code base instead of fetching it in multiple places. Once it is fetched, we reload the pusher and potentially show it in the UI. We also use this fetched info when sending chat messages and other API actions (where we today fetch the channel info every time).

id3adeye:
will re think :)

PR #185, opened 2025-11-23
added unit test when there is no never timestamp in buffer

nofuture2077:
This unit test is a demo why I believe so many frames got dropped. In a situation where no new audio frame is available everything is dropped.

It is not an endless loop - but it discards the whole buffer


nofuture2077:
I don't get why this test runs here. Fails local against your branch. Planned to deactivate it after a failed build

eerimoq:
tests are not run on PR

Issue #191, opened 2025-12-04
add support for 936p60 upload


Dustinhoefer:
Hey Eric,

can we add support for this video resolution?

"936p60 is the pixel setting that will clean up best for video with a lot of motion under 6k bitrate conditions."

Or maybe even a custom resolution setting?

<img width="2556" height="1179" alt="Image" src="https://github.com/user-attachments/assets/de9eb5a6-4695-45de-9aeb-03dac097c590" />

Kind Regards
Dustin + baumi_ms

Dustinhoefer:
> interesting =) any more information somewhere on the internet to back your claims?

https://www.reddit.com/r/obs/comments/1bzws63/why_do_some_streamers_stream_at_936p_what_is_it/

https://obsproject.com/forum/threads/936p-vs-1080p.157676/



and honestly, I have seen more 936p streams than 768p or 540p, so I think that 936p has a valid place here.


What is your opinion on adding a custom resolution field to allow even more resolutions?

PR #192, opened 2025-12-06
add 936p resolution for streaming

Dustinhoefer:
closes https://github.com/eerimoq/moblin/issues/191

this resolution is a good compromise between quality and performance.


eerimoq:
Did you test this? Because it fails to capture video from builtin camera. You probably have to capture 1080p and downscale.

Dustinhoefer:
> > Did you test this? Because it fails to capture video from builtin camera. You probably have to capture 1080p and downscale.
> 
> Nope, don't have any apple devices, sorry. Thought this would work out of the box...

I would have tried it, if I had a Mac, iPhone that was supported 😔

PR #234, opened 2026-02-12
Refactor WhipStream: ICE gathering, cleanup, and RTP improvements

Kn0ax:
Changes include:
- Wait for ICE gathering to complete before sending SDP offer (with fallback timeout)
- Replace createOffer() with getLocalDescription() for proper ICE candidates
- Use relative RTP timestamps instead of absolute for H.264
- Simplify unused delegate parameters and audio RTP packetization


eerimoq:
the PR worked for me as well, but I'll implement it differently.

The problem on main was that the wrong video timestamp was used (it was always 0), and that ffplay does only work if starting RTP timestamp at a value close to 0 (I think). The AI fixed those, but other changes in the PR were not needed, at least not for me. Also, the AI:s timestamps fixes works in the normal case, but I'm sure audio and video would go out of sync if a few audio buffers are dropped somewhere is the audio pipeline.

Either way, AI helped a lot in figuring out what was wrong, but it's fix was kinda poor imo

PR #317, opened 2026-04-09
Add configurable instant replay delay

jeromecornet:
This lets a user configure how many seconds to record after the 'Instant Replay' or 'Save Instant Replay' buttons are pressed.  Right now it's hardcoded to 5 seconds, but for my application I want to set it to 1 second (replaying the action that led to a goal for example). 

I am not a swift developer, so I tried to reproduce the patterns I saw in the code, and I tucked the delay in the Replay settings for a stream (defaulting to 5 seconds, i.e. the existing behavior)

I did not translate the strings into various languages (I'm not sure how you usually deal with it), but made them localizable.

### DIFF
--- a/Moblin/Various/Model/ModelReplay.swift
+++ b/Moblin/Various/Model/ModelReplay.swift
@@ -20,7 +20,7 @@ extension Model {
             return false
         }
         replay.isSaving = true
-        DispatchQueue.main.asyncAfter(deadline: .now() + .seconds(delay ?? 5)) {
+        DispatchQueue.main.asyncAfter(deadline: .now() + .seconds(delay ?? stream.replay.instantReplayDelay)) {
             self.replayBuffer.createFile { file in
                 DispatchQueue.main.async {
                     self.replay.isSaving = false
@@ -67,7 +67,9 @@ extension Model {
         guard replay.instantReplayCountdown == 0 else {
             return
         }
-        let savingStarted = saveReplay(start: start, delay: delay) { video in
+        let actualDelay = delay ??  stream.replay.instantReplayDelay
+        
+        let savingStarted = saveReplay(start: start, delay: actualDelay) { video in
             self.loadReplay(video: video) {
                 self.replay.isPlaying = true
                 if !self.replayPlay() {
@@ -76,7 +78,7 @@ extension Model {
             }
         }
         if savingStarted {
-            replay.instantReplayCountdown = delay ?? 6
+            replay.instantReplayCountdown = actualDelay + 1;
             instantReplayCountdownTick()
         }
     }
@@ -170,7 +172,11 @@ extension Model {
         replayEffect?.cancel()
         replayEffect = nil
     }
-
+    
+    func replayDelay(delay: Int) {
+        stream.replay.instantReplayDelay = delay
+    }
+    
     func streamReplayEnabledUpdated() {
         replayBuffer = ReplayBuffer()
         media.setReplayBuffering(enabled: stream.replay.enabled)
--- a/Moblin/Various/Model/ModelReplay.swift
+++ b/Moblin/Various/Model/ModelReplay.swift
@@ -20,23 +20,24 @@ extension Model {
             return false
         }
         replay.isSaving = true
-        DispatchQueue.main.asyncAfter(deadline: .now() + .seconds(delay ?? stream.replay.instantReplayDelay)) {
-            self.replayBuffer.createFile { file in
-                DispatchQueue.main.async {
-                    self.replay.isSaving = false
-                    guard let file else {
-                        return
+        DispatchQueue.main
+            .asyncAfter(deadline: .now() + .seconds(delay ?? stream.replay.postTriggerDuration)) {
+                self.replayBuffer.createFile { file in
+                    DispatchQueue.main.async {
+                        self.replay.isSaving = false
+                        guard let file else {
+                            return
+                        }
+                        let replaySettings = self.replaysStorage.createReplay()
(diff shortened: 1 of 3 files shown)


jeromecornet:
Yes, the slider allows selecting how much of the video to show, but the 5 seconds delay after pressing the replay button (while it's recording the action after pressing the button) is what I am trying to address.

I film water polo matches, and there is about 10-15 seconds between the time a goal is scored and the play restarts. I press the replay button right after the ball goes into the net and I want to show the 5 seconds leading to the goal.

With a 5 seconds delay post button press, it replays 10 seconds of video (5+5) starting the playback 5 seconds post goal, which means that the play has likely restarted already. And I definitely can't show 0.5x speed replays otherwise it would take 25 seconds post goal to resume the live view.

With a 1 second delay, I can show a 0.5x speed clip of 6 seconds total time, taking 12 seconds to display, resuming the live view 13 seconds after the goal (or a 1x replay even faster).

I guess I can interrupt the replay display (with the stop button) to not show the 5 seconds post goal, but the initial delay before showing the replay is also problematic.

I hope this makes sense, and I don't think the slider lets me do that.

Moblin/Various/Settings/SettingsStream.swift
```diff
@@ -763,17 +763,20 @@ struct SettingsStreamReplayStinger: Codable {
 
 class SettingsStreamReplay: Codable, ObservableObject {
     @Published var enabled: Bool = false
+    @Published var instantReplayDelay: Int = 5
```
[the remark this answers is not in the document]
jeromecornet: 👍

Moblin/View/Settings/Streams/Stream/Replay/StreamReplaySettingsView.swift
```diff
@@ -137,6 +137,23 @@ struct StreamReplaySettingsView: View {
...
+                                localized: "Seconds to record after the Instant replay button is pressed"
+                            ),
+                        ],
+                        keyboardType: .numbersAndPunctuation
+                    ) {
+                       model.replayDelay(delay: Int($0) ?? 5)
+                    }
+                } label: {
+                    TextItemView(name: "Instant replay delay", value: "\(String($replay.instantReplayDelay.wrappedValue)) s")
+                }
```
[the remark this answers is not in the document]
jeromecornet: Done

jeromecornet:
Updated the property name, default to 2 seconds and used a slider. Would it make sense to making the delay configurable from 2 to 10 seconds (rather than 0 to 5) to avoid setting it to a value that may lead to an unexpected behavior ?

I also fixed the code style (sorry I forgot to do that last time)

eerimoq:
let's start with 0 to 5, and test it

Moblin/Various/Model/ModelReplay.swift
```diff
@@ -67,7 +68,9 @@ extension Model {
         guard replay.instantReplayCountdown == 0 else {
             return
         }
-        let savingStarted = saveReplay(start: start, delay: delay) { video in
+        let actualDelay = delay ?? stream.replay.postTriggerDuration
+
```
eerimoq: no blank lines within function bodies

Moblin/Various/Model/ModelReplay.swift
```diff
@@ -171,6 +174,10 @@ extension Model {
         replayEffect = nil
     }
 
+    func replayDelay(delay: Int) {
+        stream.replay.postTriggerDuration = delay
+    }
+
```
eerimoq: remove this function

jeromecornet: whoops, it was left over from a first attempt at hooking it into the replay screen rather than settings

Moblin/View/Settings/Streams/Stream/Replay/StreamReplaySettingsView.swift
```diff
@@ -106,6 +106,14 @@ struct StreamReplaySettingsView: View {
     let stream: SettingsStream
     @ObservedObject var replay: SettingsStreamReplay
 
+    private func submitInstantReplayDelay(value: Float) {
+        model.replayDelay(delay: Int(value))
+    }
+
+    private func formatInstantReplayDelay(value: Float) -> String {
+        return String(Int(value)) + " s"
```
eerimoq: `"\(Int(value)) s"` instead

Moblin/View/Settings/Streams/Stream/Replay/StreamReplaySettingsView.swift
```diff
@@ -137,6 +145,19 @@ struct StreamReplaySettingsView: View {
...
+                               minimum: 0,
+                               maximum: 5,
+                               step: 1,
+                               onSubmit: submitInstantReplayDelay,
+                               width: 70,
+                               format: formatInstantReplayDelay)
+                } header: {
+                    Text("Instant replay delay")
+                } footer: {
+                    Text("Seconds to record after the Instant replay button is pressed")
```
eerimoq: should probably not be called instant replay

jeromecornet: How about this ?
```swift
                } header: {
                    Text("Replay save delay")
                } footer: {
                    Text("Seconds to record after the Instant replay/Save replay button is pressed")
                }
```

eerimoq:
I might replace the slider with a picker after merging. not sure what is best

eerimoq:
it's in test flight now

PR #321, opened 2026-04-24
Merge new DeepLink streams by name 

jkaretsky:
Every time a DeepLink is imported, a new stream is added to the streams list. This can get impractical if a user is frequently adding new streams with DeepLinks. 

This change looks to see if any of the imported streams match an existing stream by name, and if so, it updates their settings to the extent specified in the DeepLink. This allows DeepLinks to update existing streams without polluting the streams list.


eerimoq:
image
image

eerimoq:
feel free to join the discussion =)

eerimoq:
a dialog asking if the stream should be replaced or created should be good, as spillmaker suggested in the chat above.

PR #323, opened 2026-04-27
Hide scoreboard clock when it reaches 0:00

jeromecornet:
In the modular scoreboard, it is impossible to display the period without displaying the clock (I'm using the side by side view with info box).

In games with a shot clock (like water polo or hockey) where the clock doesn't run continuously,  I don't want to display a clock that is not the actual game clock in the field, but knowing which period we are in is still useful (and I can update it with the remote control).

This patch simply hides the clock display when it reaches 0:00, so in sports where you don't want to see the clock but want to see the period, simply set the clock to 0:00 and it will disappear.

### DIFF
--- a/Moblin/VideoEffects/Scoreboard/ScoreboardEffectModularView.swift
+++ b/Moblin/VideoEffects/Scoreboard/ScoreboardEffectModularView.swift
@@ -156,10 +156,12 @@ struct ScoreboardEffectModularView: View {
                                 .font(.system(size: fontSize() * 0.6))
                                 .minimumScaleFactor(0.1)
                         }
-                        Text(config.global.timer)
-                            .font(.system(size: fontSize() * 0.9))
-                            .monospacedDigit()
-                            .minimumScaleFactor(0.1)
+                        if config.global.timer != "0:00" {
+                            Text(config.global.timer)
+                                .font(.system(size: fontSize() * 0.9))
+                                .monospacedDigit()
+                                .minimumScaleFactor(0.1)
+                        }
                     }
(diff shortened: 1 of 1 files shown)


eerimoq:
hmm did you consider adding a toggle to hide/show the clock?

jeromecornet:
> hmm did you consider adding a toggle to hide/show the clock?

I thought it would add unnecessary complexity, the web remote is already quite cluttered, and a 0:00 clock showing didn't seem to make sense anyway (maybe in some sports you want to know that it ran out?) so I figured I would quickly hack this.

There is already some inconsistencies between the various views:
- the clock shows in side by side, but not stacked or stacked history
- the period shows in side by side and stacked history, but not stacked

Rather than a toggle, maybe this should be a new view ? like side by side with no clock, updating the remote so we remove the clock management from the remote UI ?

eerimoq:
@scyc314 is the brain behind the scoreboard. Maybe he has some input.

hiding the clock when it reaches 0:00 will not be the expected behavior for many sports, I'm sure

jeromecornet:
That said for sports using stop-time clocks, even though the clock is a part of the game, it's virtually impossible to keep the scoreboard clock in sync with the IRL game clock, so I think it makes sense to have a way to hide the clock.

TBH the status quo is that I don't show the period if I am not showing the clock, which is fine by me. But I would welcome ideas on how to make the scoreboard more flexible without having to manually think of every possible sport's idiosyncrasies.

eerimoq:
I think adding a Clock toggle in settings will work as a first step

jeromecornet:
closing in favor of https://github.com/eerimoq/moblin/pull/324

PR #333, opened 2026-05-14
Add support for Dji Pocket 4

ducker:
Support for Dji Pocket 4


eerimoq:
how did you test it?

ducker:
I have on my own Dji Action 4 and Dji Pocket 4. I was a bit surprised that they changed the format and was not compatible with existing implementation so I did some research about format they use and then reverse engineered the BLE messages from Mimo application to set the correct params. I verified if the output and bitrate matches one selected by the user and of course if video and audio are correct both in preview and when transmitting to Srt server.

tentenlee100:
I need this

eerimoq:
merged locally

PR #334, opened 2026-05-14
Allow update URL for Media source

ducker:
Allows updating the FFmpeg source URL remotely via the WebSocket API.

### DIFF
--- a/Moblin/Obs/ObsWebSocket.swift
+++ b/Moblin/Obs/ObsWebSocket.swift
@@ -243,12 +243,6 @@ struct SetCurrentProgramSceneRequest: Codable {
 
 struct SetItemUrlInputSettings: Encodable {
     let input: String
-    let isLocalFile: Bool
-
-    enum CodingKeys: String, CodingKey {
-        case input
-        case isLocalFile = "is_local_file"
-    }
 }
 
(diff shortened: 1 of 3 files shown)


eerimoq:
show some screenshots and describe the use case

ducker:
I'm taking care of an obs configuration for some IRL streamer who often shares the pov with others. I have separate scene for that purpose so it can be easily switched but to update the srt URL I need to use Remote Desktop and manually change that value in obs. It would be quite handy if I can do this through Moblin app instead.

IMG_3667
IMG_3668
IMG_3669

eerimoq:
cool. usually in moblin we don't use alerts to enter some setting, but navigation links to subpages. so let's do this similar to how the rest of the app does it

ducker:
Updated UI flow
IMG_3670
IMG_3671
IMG_3672
IMG_3673

eerimoq:
I think we should fetch the "input" field value so we can present the current value in the UI so you know what you are changing from

eerimoq:
It's probably a good idea not to mention ffmpeg at all, but just as in OBS, call it Media source, and show/modify media source settings

ducker:
IMG_3674
IMG_3676

eerimoq:
Merged it locally

PR #336, opened 2026-05-16
Allow to use current wifi settings if location permission is not yet given

ducker:
Current implementation autofills the field only when location permission is already given. I changed that and added a button which ask for permission if not yet given and allow to fill textfield with WiFi SSID. It also preselects stream url.

### DIFF
--- a/Moblin/View/Settings/DjiDevices/DjiDeviceSettingsView.swift
+++ b/Moblin/View/Settings/DjiDevices/DjiDeviceSettingsView.swift
@@ -1,5 +1,63 @@
+import CoreLocation
 import NetworkExtension
 import SwiftUI
+
+private final class WiFiSsidFetcher: NSObject, CLLocationManagerDelegate, ObservableObject {
+    enum Failure: Error {
+        case notConnected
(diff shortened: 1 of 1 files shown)


eerimoq:
One feature per PR. Looks like a lot of code for a smallish improvement? Any chance you can refactor it to make it easier to understand?

ducker:
In my opinion it's one feature. Added button that allow to ask user for location permission if not granted prior and fills the configuration details. Do you have some idea how to split it into smaller pieces?

Most of the added lines comes from localisation file as there are some toasts information for the user to better understand the process.

eerimoq:
automatically filling in the password seems like a good first PR, but instead of doing it when an SSID has been discovered, do it in the SSID submit callback. I'm not yet sure all other changes should be merged. have to think about it. last week we removed some of the auto selecting as it was confusing to streamers, not knowing why their settings randomly changed. we may add more magic again, but let's not rush it

eerimoq:
we might add all of it soonish. had some feedback at twitchcon =)

PR #338, opened 2026-05-22
Add stealth mode functionality to remote control

datagutt:
I only tested web because I only have 1 apple device


eerimoq:
one one device is needed. it can control itself

eerimoq:
I merged it locally

PR #352, opened 2026-06-01
Option to merge new DeepLink streams with existing streams with same name

jkaretsky:
Incremental changes to #321 so new DeepLink streams which have matching names with existing streams have the option to be modified in place without having to add a new stream


eerimoq:
Merged locally. Thanks!

PR #353, opened 2026-06-02
Support Kick badges_v2 (level badge)

id3adeye:


Moblin/StreamingPlatforms/Kick/KickPusher.swift
```diff
@@ -59,9 +59,15 @@ private class KickBadges {
     }
 }
 
+private struct BadgeV2: Decodable {
+    var image_url: String?
+    var selected: Bool?
```
eerimoq: are these really optional?

id3adeye: fixed :)

PR #355, opened 2026-06-03
feat: support for osmo action 6

nofuture2077:
Implement initial OA6 protocol support based on Bluetooth capture analysis

Reverse-engineered the OA6 protocol from a Bluetooth capture of the Mimo app (simon_offiziell) using Claude 4.7 and implemented a minimal change set.

Testing with an Osmo Action 6 shows that lag issues in certain stabilization modes are resolved. Streaming can now be started and behaves similarly to the existing OA4 flow.

Further testing is needed to validate stability and uncover any remaining issues.

### DIFF
--- a/Moblin/Integrations/Dji/DjiDevice/DjiDevice.swift
+++ b/Moblin/Integrations/Dji/DjiDevice/DjiDevice.swift
@@ -321,7 +321,10 @@ extension DjiDevice: CBPeripheralDelegate {
         switch model {
         case .osmoAction2, .osmoAction3:
             sendStartStreaming()
-        case .osmoAction4:
+        case .osmoAction4, .osmoAction6:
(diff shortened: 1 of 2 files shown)


nofuture2077:
Just for context: The OA6 is a crazy good streaming action cam. Used it 9 days in a row (40+ streaming hours) on a bike stream through Netherlands. No connection issue / no microphone issues / no lags. Streamed all of it with moblin - just used the Mimo App to start the action cam stream targeting the RTMP of moblin. This PR mimics the mimo app and makes it a little more convenient to use for streamers.

vismajor1:
I agree. I’ve also used the OA6 for a few streams. Once I used the built-in microphone, and for most streams I connected two DJI Mic Minis directly to the camera. It works really well, and I like the image quality in low light, too. I also set up the connection using the Mimo app. Simon_offiziell

PR #361, opened 2026-06-09
Browser widget: fix UTF-8 decoding in the JavaScript bridge

thecoolrhyno:
Messages are delivered to a Browser widget's web page base64-encoded and
decoded with `atob()`, which returns a *binary string* (one byte per
character). Passing that directly to `JSON.parse()` makes multi-byte UTF-8
sequences be read as Latin-1, so any non-ASCII text arrives corrupted — e.g. a
chat message or speech-to-text transcript containing "música" reaches the page
as "mÃºsica".

This reconstructs the bytes from the binary string and decodes them as UTF-8
with `TextDecoder` before parsing:

```js
const bytes = Uint8Array.from(message, (c) => c.charCodeAt(0));
const json = new TextDecoder().decode(bytes);
this.onmessage(JSON.parse(json).message.data);
```

One-line behavior change in `moblin.js`; affects every topic delivered to the
web page (chat included). ASCII content is unaffected.

### DIFF
--- a/Moblin/VideoEffects/Browser/moblin.js
+++ b/Moblin/VideoEffects/Browser/moblin.js
@@ -16,7 +16,11 @@ class Moblin {
 
   handleMessage(message) {
     if (this.onmessage) {
-      this.onmessage(JSON.parse(message).message.data);
+      // `message` is a binary string (one byte per char) from atob(); decode it
+      // back to UTF-8 so non-ASCII characters (accents, etc.) survive.
+      const bytes = Uint8Array.from(message, (c) => c.charCodeAt(0));
+      const json = new TextDecoder().decode(bytes);
+      this.onmessage(JSON.parse(json).message.data);
     }
   }
 
--- a/Moblin/VideoEffects/Browser/moblin.js
+++ b/Moblin/VideoEffects/Browser/moblin.js
@@ -16,10 +16,9 @@ class Moblin {
 
   handleMessage(message) {
     if (this.onmessage) {
-      // `message` is a binary string (one byte per char) from atob(); decode it
-      // back to UTF-8 so non-ASCII characters (accents, etc.) survive.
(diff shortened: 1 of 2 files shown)


eerimoq:
why not use `Uint8Array.fromBase64()`?

thecoolrhyno-gutz:
> why not use `Uint8Array.fromBase64()`?

would love to, but fromBase64() is Safari 18.4+ and Moblin targets iOS 16.4 — it'd throw on older devices (widget uses the system WebKit). atob + TextDecoder is boring but works on all supported versions for ~the same code. happy to switch if you wanna bump the min iOS 👍

eerimoq:
it's ok not to support every feature on old devices

thecoolrhyno-gutz:
> it's ok not to support every feature on old devices

Done — moved to Uint8Array.fromBase64(), much cleaner. Thanks!

Moblin/VideoEffects/Browser/moblin.js
```diff
@@ -16,7 +16,10 @@ class Moblin {
 
   handleMessage(message) {
     if (this.onmessage) {
-      this.onmessage(JSON.parse(message).message.data);
+      // `message` is base64; decode straight to UTF-8 so non-ASCII characters
+      // (accents, emoji, etc.) survive instead of being read as Latin-1.
+      const json = new TextDecoder().decode(Uint8Array.fromBase64(message));
```
eerimoq: remove the comment and create the text decoder once instead of once per message.

thecoolrhyno: Done — single TextDecoder in the constructor now, and comment removed

eerimoq:
I just tested it and it doesn't work =( Probably some bug that is easy to fix. Please have a look at it.

eerimoq:
hmm, now it worked

PR #362, opened 2026-06-09
Browser widget: expose speech-to-text transcript to web widgets

thecoolrhyno:
## What

Adds a `speechToText` subscribe topic to the Browser widget JavaScript API,
mirroring the existing `chat` topic. A subscribed web page receives the
recognized transcript — the same text the subtitles feature already produces —
via `moblin.onmessage`:

```js
moblin.subscribe({ speechToText: {} });
moblin.onmessage = (message) => {
  if (message.speechToText !== undefined) {
    console.log(message.speechToText.position, message.speechToText.text);
  }
  if (message.speechToTextClear !== undefined) {
    // Transcript was cleared after silence.
  }
};
```

No new data is exposed: it is the same transcript the Subtitles feature and
the speech-to-text Alert trigger already use, only routed to the widget. It is
gated by the existing "Moblin access" setting plus a new per-widget "Speech to
text" toggle.

## Design

- **Per-widget toggle, no hot-path change.** Following your feedback about
  efficiency, the trigger is a per-widget Browser setting (like
  `needsSubtitles` on text/alerts widgets): `isSpeechToTextNeeded()` is decided
  from settings at scene update time, so the recognizer starts/stops without
  any periodic re-evaluation. The periodic timers are untouched.
- **Payload mirrors the internal delegate.** Messages carry `text` (the
  sliding transcript window) and `position` (absolute offset of its first
  character), exactly what `speechToTextPartialResult(position:text:)`
  delivers, so web pages can dedupe partial updates the same way the alerts
  widget does internally. A `speechToTextClear` message is sent when the
  transcript clears (silence timeout and recognizer stop), matching what text
  widgets get via `clearSubtitles()`.
- **Per-widget delivery gate.** `BrowserEffect` only forwards the transcript
  when its own widget's toggle is on — with two browser widgets, a page whose
  widget has the toggle off never receives speech, even while the recognizer
  runs for the other one. The toggle is the user's per-page consent for
  receiving their speech.
- **Settings compatibility:** the new field decodes with `false` as default,
  old configs are unaffected.

## Use cases

Custom caption overlays, "say a phrase to trigger an animation", voice-driven
widgets — same idea as the existing speech-to-text Alert trigger, available to
any web widget.

(My own use case: a voice-assistant widget. On iOS, `getUserMedia` inside the
widget's webview fights the broadcast's `AVAudioSession` — grabbing the mic
switches the live input route. Riding on Moblin's existing recognizer avoids
that entirely.)

## Testing

- `make style-check`, `make lint` (swiftlint --strict), `make spell-check` and
  the full CI workflow (including the Xcode build) green.
- Tested on device: a Browser widget pointing at a test page subscribes to
  `speechToText`; with "Moblin access" + "Speech to text" on, the page receives
  the live transcript (and `speechToTextClear` after silence) while streaming,
  with no microphone access from the webview.

### DIFF
--- a/Moblin/Various/Model/ModelScene.swift
+++ b/Moblin/Various/Model/ModelScene.swift
@@ -1054,7 +1054,7 @@ extension Model {
             case .text:
                 addSceneTextEffects(sceneWidget, widget, &effects, &needsSpeechToText)
             case .browser:
-                addSceneBrowserEffects(sceneWidget, widget, scene, &effects)
+                addSceneBrowserEffects(sceneWidget, widget, scene, &effects, &needsSpeechToText)
             case .crop:
                 addSceneCropEffects(widget, scene, &effects)
             case .map:
@@ -1122,7 +1122,8 @@ extension Model {
         _ sceneWidget: SettingsSceneWidget,
         _ widget: SettingsWidget,
         _ scene: SettingsScene,
-        _ effects: inout [VideoEffect]
+        _ effects: inout [VideoEffect],
+        _ needsSpeechToText: inout Bool
     ) {
         guard let effect = browserEffects[widget.id], !effects.contains(effect) else {
             return
@@ -1132,6 +1133,9 @@ extension Model {
             crops: findWidgetCrops(scene: scene, sourceWidgetId: widget.id)
         )
         effects.append(effect)
+        if widget.browser.moblinAccess, widget.browser.speechToText {
+            needsSpeechToText = true
+        }
     }
 
     private func addSceneCropEffects(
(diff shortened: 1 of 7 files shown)


eerimoq:
why doesn't the builtin alert triggered by speech to text work in your case?

thecoolrhyno-gutz:
> why doesn't the builtin alert triggered by speech to text work in your case?

Yeah, the built-in alert works great for a fixed keyword → alert. Our case is a bit different though
  — we're building a little voice assistant (we call it Rino) that lives in a browser widget and
  actually helps run the stream: you say a wake word, then ask it stuff like "pause the alerts", "skip
  this song", "how many viewers right now?", and it figures out the intent and acts on it.
  
  For that we need the raw transcript to reach the widget itself, which the alert path doesn't expose —
  it just fires an alert on a preset phrase. So this PR is really just the minimal bridge: hand the
  on-device STT to browser widgets (toggleable per widget). Ours is one use, but it opens the door for
  any widget that wants to do something with speech.

  Happy to tweak the API if you'd rather surface the transcript a different way!

eerimoq:
Merged locally.

PR #366, opened 2026-06-11
Add per-source audio offset setting for all ingest types

nofuture2077:
Adds a -2000..+2000 ms audio offset slider to RTMP, SRTLA, RIST, SRT client, WHIP, and WHEP stream settings. The offset is applied as a static correction on top of the DriftTracker in BufferedAudio, so automatic sync still works while the user can nudge audio early or late to fix A/V async.


nofuture2077:
I will test this tomorrow with the cams i have on hand. This is a draft

nofuture2077:
This works as intended. The slider is available per source and can be updated "live". Uses your existing audio drift mechanism. 

Where I am unsure: +- 2s range could be a bit extensive - On the other hand some people experience audio async of over a second.

PR #369, opened 2026-06-12
Interactive Widget Overlay (Drag, Drop, and Resize)

paraisospelosul:
This PR introduces an interactive mode to edit, drag, resize, and lock widgets directly on the camera preview without cluttering the core video pipeline.

Architectural Approach
We avoided injecting complex touch handlers into the core Metal / CoreImage pipeline. Instead, we built a 100% declarative SwiftUI overlay (InteractiveWidgetOverlayView.swift) that floats invisibly above the camera preview. It mirrors the exact bounding boxes of the active widgets and uses SwiftUI DragGesture and MagnificationGesture to rewrite the coordinates (SettingsWidgetLayout) in real-time.

Key Changes
1. The Overlay View (InteractiveWidgetOverlayView.swift)
A new standalone SwiftUI component that contains all the layout dragging, pinching, snapping, and HUD logic.
Computes exact dimensions using getWidgetDimensions to ensure the invisible hitboxes perfectly match the widget's real bounds.
Text Widget Fixes: Text widgets don't scale using layout.size. The overlay correctly identifies .text widgets, estimates their bounding box based on fontSizeFloat, and intercepts pinch-to-zoom gestures to inject changes directly into fontSize instead of the layout bounds, avoiding the "giant hitbox" bug.
Smart Locking: Respects Moblin's existing positioningLock. A locked widget can still be selected via tap (to reveal the unlock HUD) but prevents drag and pinch gestures.
2. Model Hooks (ModelScene.swift)
Added updateWidgetLayoutDirectly() to allow the overlay to sync layout changes in real-time during a 60fps drag without constantly writing to the persistent database, saving CPU and avoiding stutter.
3. Integration (MainView.swift & Settings)
Injected InteractiveWidgetOverlayView as an overlay inside MainView.swift.
Added an "Edit layout on screen" toggle inside QuickButtonSceneWidgetsView.swift so the user can easily toggle the feature on/off.
Why this approach?
It is completely non-invasive. If the edit mode is toggled off, the overlay isn't even rendered. It leverages the existing SettingsWidgetLayout and doesn't require rewriting any of the complex widget rendering logic in ImageEffect.swift or TextEffect.swift.

### DIFF
--- a/Moblin/Various/Model/Model.swift
+++ b/Moblin/Various/Model/Model.swift
@@ -401,6 +401,8 @@ final class Model: NSObject, ObservableObject, @unchecked Sendable {
     var pendingStreamImportCollisionTitle: String = ""
     @Published var showDrawOnStream = false
     @Published var showLocalOverlays = true
+    @Published var editWidgetsMode = false
+    @Published var selectedWidgetForInteraction: WidgetInScene?
     @Published var showBrowser = false
     @Published var showNavigation = false
     @Published var webBrowserUrl: String = ""
@@ -1155,6 +1157,7 @@ final class Model: NSObject, ObservableObject, @unchecked Sendable {
         bonding.statisticsFormatter.setNetworkInterfaceNames(database.networkInterfaceNames)
         reloadTeslaVehicle()
         updateQuickButtonStates()
+        editWidgetsMode = database.quickButtons.first(where: { $0.type == .editWidgets })?.isOn ?? false
         setQuickButton(type: .blurFaces, isOn: database.face.blurFaces)
         setQuickButton(type: .blurText, isOn: database.face.blurText)
         setQuickButton(type: .privacy, isOn: database.face.blurBackground)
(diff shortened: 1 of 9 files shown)


Kn0ax:
You included all your previous commits, which isn’t the ideal approach for PRs. 
Please only include the relevant commit(s) in the PR :)

eerimoq:
Just some feedback with some ideas of improvements/simplifications for a first release.

- Remove the buttons that shows next to the widget. the up, down, lock and delete buttons.
- Remove the "Edit layout on screen" toggle, and instead make widgets editable on screen when the "Scene widgets" panel is open.
- What are the Clickable toggles used for?
- Never make position or size numbers negative, as currently happens when moving a widget partly outside the video to the top or left.
- Do not show a "rectangle" indicating the widget size if the widget does not have a layout size setting. Show something else to "select" it, maybe a move icon or something.
- Style the widget rectangle to look the same as the polygon in the mask widget effect. It should still be a rectangle, just color and circles should change.

PR #371, opened 2026-06-14
feat: visual checkmark badges (NewFeatureIndicator) to highlight newl…

paraisospelosul:
### Performance and Versioning Details

- **Zero Performance Overhead**: The indicator system has virtually zero CPU and memory impact. The version lookups are done using a lightweight static dictionary with `O(1)` complexity, running in microseconds only when settings screens are opened. Since we use SwiftUI conditional rendering, the badges are not even loaded or drawn in the UI when they are inactive.
- **Automatic Version Management**: On each new Moblin release, the `NewFeatureManager` compares the current app version with the mapped list of features.
  - It highlights only the new features introduced in the current active version.
  - Once the user updates to a newer release, older badges automatically disappear from the UI without requiring any manual code removal or cleanup.


eerimoq:
Looks like it will be too much work to keep this updated, sorry.

PR #372, opened 2026-06-14
Add region, area, and suburb text widget variables

### LINKED ISSUE #363: Additional Location Variables for Text Widgets
First of all, thank you for the excellent work on Moblin. I recently switched from another streaming app and especially appreciate the image quality, adaptive bitrate handling and travel-streaming features.

For travel and IRL streams, I would like to request additional location variables for Text Widgets.

Currently available variables such as:

{countryFlag}
{city}

are already very useful.

However, many travel streamers would like to display more geographical context than just the city name. Since administrative structures differ between countries (county, municipality, district, region, etc.), I would suggest using generic location levels instead of country-specific names.

Suggested variables:

{region}
{area}
{city}
{suburb}

and optionally:

{fullLocation}

Examples:

Germany:
{region} = North Rhine-Westphalia
{area} = Lippe
{city} = Schieder-Schwalenberg
{suburb} = Glashütte

Norway:
{region} = Vestland
{area} = Aurland
{city} = Flåm
{suburb} = Fretheim

This would allow streamers to build custom overlays such as:

🇩🇪 Lippe, Schieder-Schwalenberg
or
🇳🇴 Vestland, Flåm

while still keeping the implementation independent of country-specific administrative systems.

For users who prefer a simple solution, an additional variable such as:
{fullLocation}
could provide a preformatted location string.
This would be especially useful for travel streams, road trips, hiking streams, tourism content and GPS-based overlays.

Similar to the already existing {city} variable, but exposing more of the location hierarchy that Moblin already resolves internally.

Thank you for considering this feature

mvanhorn:
## Summary
Adds `{region}`, `{area}`, and `{suburb}` text-widget variables, mirroring the existing `{city}`/`{state}` variables. These map onto the `CLPlacemark` fields the app already resolves via `GeographyManager`:

- `{region}` -> `administrativeArea` (a travel-friendly alias of the existing `{state}`)
- `{area}` -> `subAdministrativeArea` (county/district) -- not previously exposed
- `{suburb}` -> `subLocality` (neighborhood/suburb) -- not previously exposed

`{region}` and `{state}` intentionally both map to `administrativeArea`; `{state}` is kept for backward compatibility while `{region}` is the country-agnostic name requested in the issue.

## Why this matters
Travel and IRL streamers want to surface more geographic hierarchy than just `{city}`, especially in countries where "state" is the wrong label. The placemark already carries county/district and neighborhood data, so this just wires the existing values through the text pipeline. https://github.com/eerimoq/moblin/issues/363

## What changed
The three variables are wired through every stage of the text pipeline exactly like `{city}`:
- `TextFormatStringLoader.swift`: new enum cases, parse branches, and `isLocationVariable()` entries.
- `TextEffect.swift`: new `TextEffectStats` fields.
- `TextEffectFormatter.swift`: render cases plus `formatRegion`/`formatArea`/`formatSuburb` (`?? "-"` fallback, matching `formatCity`).
- `ModelScene.swift`: populate from the placemark and include the new cases in `updateNeedsGeography`.
- `RemoteControl.swift`: new fields on `RemoteControlRemoteSceneDataTextStats` plus `init(stats:)` and `toStats()` pass-through, so remote-scene streamers get the values too.
- `WidgetTextSettingsView.swift` + `Localizable.xcstrings`: list the new variables in the Text widget variable picker.

## Testing
Added parse and render tests to `MoblinTests/TextEffectSuite.swift` (parse of `{region}`/`{area}`/`{suburb}`, mixed `{area}, {city}`, present/absent render, and `isLocationVariable()`), and extended `createStats()` with the new fields.

I could not run the test suite headlessly: Moblin builds only through Xcode and requires a configured `Config/User.xcconfig` (DEVELOPMENT_TEAM / bundle identifier) plus full SPM resolution, neither of which is available in my environment. The change is a line-for-line mirror of the existing `{city}`/`{state}` implementation across all pipeline stages. Please run `Command + U` in Xcode to confirm.

Fixes #363

### DIFF
--- a/Moblin/Various/Model/ModelScene.swift
+++ b/Moblin/Various/Model/ModelScene.swift
@@ -1536,6 +1536,9 @@ extension Model {
                 countryFlag: emojiFlag(countryCode: placemark?.isoCountryCode),
                 state: placemark?.administrativeArea,
                 city: placemark?.locality,
+                region: placemark?.administrativeArea,
+                area: placemark?.subAdministrativeArea,
+                suburb: placemark?.subLocality,
                 muted: isMuteOn,
                 heartRates: heartRates,
                 activeEnergyBurned: workoutActiveEnergyBurned,
@@ -1763,7 +1766,7 @@ extension Model {
     private func updateNeedsGeography(_ text: SettingsWidgetText, _ parts: [TextFormatPart]) {
         text.needsGeography = parts.contains(where: {
             switch $0 {
-            case .country, .countryFlag, .state, .city:
+            case .country, .countryFlag, .state, .city, .region, .area, .suburb:
                 true
             default:
                 false
--- a/Moblin/RemoteControl/RemoteControl.swift
+++ b/Moblin/RemoteControl/RemoteControl.swift
@@ -555,6 +555,9 @@ struct RemoteControlRemoteSceneDataTextStats: Codable {
     var countryFlag: String?
     var state: String?
     var city: String?
+    var region: String?
+    var area: String?
+    var suburb: String?
     var muted: Bool
     var heartRates: [String: Int?]
     var activeEnergyBurned: Int?
@@ -594,6 +597,9 @@ struct RemoteControlRemoteSceneDataTextStats: Codable {
         countryFlag = stats.countryFlag
         state = stats.state
         city = stats.city
+        region = stats.region
+        area = stats.area
+        suburb = stats.suburb
         muted = stats.muted
         heartRates = stats.heartRates
         activeEnergyBurned = stats.activeEnergyBurned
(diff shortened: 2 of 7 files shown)


Moblin/Various/Model/ModelScene.swift
```diff
@@ -1536,6 +1536,9 @@ extension Model {
                 countryFlag: emojiFlag(countryCode: placemark?.isoCountryCode),
                 state: placemark?.administrativeArea,
                 city: placemark?.locality,
+                region: placemark?.administrativeArea,
```
eerimoq, line 1539: why not only pass `placemark?.administrativeArea` once to the constructor?

eerimoq:
Looks good in general!

eerimoq:
meged locally. removed state alias for now and some renaming for better or worse =)

Issue #375, opened 2026-06-17
Expose Auto Scene Switcher shuffle/random option in Remote Control API


mqtt12:
Hi,

I’m using Moblin’s Remote Control API together with a custom web panel. I can select/enable/disable an Auto Scene Switcher via remote control using:

```json
{
  "setAutoSceneSwitcher": {
    "id": "..."
  }
}
```

That works fine.

However, I would also like to read and control the “Random” / “Shuffle” option of an Auto Scene Switcher from the Remote Control API.

I found that Moblin stores this option internally as:

```swift
@Published var shuffle: Bool = false
```

in:

```text
Moblin/Various/Settings/SettingsScene.swift
class SettingsAutoSceneSwitcher
```

and it is used here:

```swift
if autoSwitcher.shuffle {
    autoSceneSwitcher.sceneIds.shuffle()
}
```

in:

```text
Moblin/Various/Model/ModelAutoSceneSwitcher.swift
```

But the Remote Control API currently seems to expose Auto Scene Switchers only with `id` and `name`.

For example, `getSettings` returns:

```json
{
  "autoSceneSwitchers": [
    {
      "id": "...",
      "name": "Test"
    }
  ]
}
```

There is no `shuffle` / `random` field.

I also checked the web remote frontend and it seems to only support:

```ts
setAutoSceneSwitcher(id: string | null): void {
  this.sendRequest({ setAutoSceneSwitcher: { id } });
}
```

So currently it looks like Remote Control can select an Auto Scene Switcher, but cannot read or change its shuffle/random mode.

Would it be possible to add support for this?

Suggested API extension:

```json
{
  "setAutoSceneSwitcherShuffle": {
    "id": "AUTO_SCENE_SWITCHER_UUID",
    "shuffle": true
  }
}
```

And maybe extend `getSettings` so each Auto Scene Switcher includes:

```json
{
  "id": "...",
  "name": "Test",
  "shuffle": false
}
```

Optionally it would also be useful if the scene list and timings of each Auto Scene Switcher could be exposed, for example:

```json
{
  "id": "...",
  "name": "Test",
  "shuffle": false,
  "scenes": [
    {
      "id": "...",
      "name": "Rot2",
      "time": 15
    }
  ]
}
```

Use case: external remote panels for IRL streaming should be able to show and control whether an Auto Scene Switcher runs sequentially or randomly, without needing to change it directly inside the Moblin app.

Thanks a lot for Moblin — it’s a great tool.

[...]

PR #379, opened 2026-06-21
feat: expose Auto Scene Switcher shuffle in Remote Control API

mvanhorn:
## Summary

Exposes the Auto Scene Switcher "Shuffle" (random) mode through the Remote Control API so external remote panels can both read and toggle it.

- `getSettings` now returns a `shuffle` field on each entry of `autoSceneSwitchers`, sourced from `SettingsAutoSceneSwitcher.shuffle`.
- Adds a new request, `setAutoSceneSwitcherShuffle(id:shuffle:)`, alongside the existing `setAutoSceneSwitcher(id:)`, wired through the streamer dispatch, the web dispatch, and the assistant client.
- The handler updates the stored `shuffle` flag for the matching switcher. If the changed switcher is the one currently running, it drops the precomputed scene order so the next switch rebuilds it with the new setting, without disrupting the active scene.
- Mirrors the new request in the web frontend client (`WebRemoteControlFrontend/src/utils.ts`) with a `setAutoSceneSwitcherShuffle(id, shuffle)` method and a typed `shuffle?` field on the settings switcher item.

The `shuffle` field on the settings DTO is optional, so a `getSettings` response that omits it decodes to `nil` (treated as off) rather than failing — keeping older and newer clients compatible.

## Why this matters

The Remote Control API could select/enable/disable an Auto Scene Switcher but had no way to inspect or change its Shuffle mode, even though the flag already exists internally on `SettingsAutoSceneSwitcher` and is consumed by the switch loop. IRL streamers driving Moblin from an external remote panel had no way to flip Shuffle remotely. This adds that read/write path. The maintainer invited a PR on the issue.

## Testing

- `swiftformat` over the touched Swift files reports no changes (style-clean).
- Verified the new request is dispatched in both the streamer (`RemoteControlStreamer`) and web (`RemoteControlWeb`) paths and routed to the model handler, matching the existing `setAutoSceneSwitcher` wiring across all four layers.
- Unknown/missing `id` is a no-op (guarded lookup, no crash). Toggling Shuffle on the active switcher rebuilds the scene order on the next switch rather than mid-scene.
- A full Xcode build/test run was not performed in this environment (iOS app build requires signing/simulator setup); the change is additive and follows the existing pattern exactly.

Refs #375


eerimoq:
If someone could make AI less verbose :)

Moblin/Various/Model/ModelAutoSceneSwitcher.swift
```diff
@@ -16,6 +16,19 @@ extension Model {
...
+        guard let switcher = database.autoSceneSwitchers.switchers.first(where: { $0.id == id }) else {
+            return
+        }
+        switcher.shuffle = shuffle
+        // If the changed switcher is currently running, drop the precomputed scene
+        // order so the next switch rebuilds it with the new shuffle setting. This
+        // avoids disrupting the current scene while the loop keeps running.
+        if autoSceneSwitcher.currentSwitcherId == id {
+            autoSceneSwitcher.sceneIds.removeAll()
+        }
```
eerimoq, line 29: let's remove this to have same logic as when configuring locally

Moblin/RemoteControl/RemoteControl.swift
```diff
@@ -773,6 +774,7 @@ struct RemoteControlSettingsScene: Codable, Identifiable {
 struct RemoteControlSettingsAutoSceneSwitcher: Codable, Identifiable {
     let id: UUID
     let name: String
+    var shuffle: Bool?
```
eerimoq, line 777: is it important that this is optional?

PR #390, opened 2026-06-26
feat: add customizable looping mute audio feedback and localized overlay

paraisospelosul:
Adds customizable looping mute audio feedback with a localized visual overlay ("Mute"/"Mudo") in the center of the screen when muted.
Key features:

Plays a customizable sound in an infinite loop while muted, stopping immediately when unmuted.
Users can choose from bundled alerts or import custom audio files directly from their device storage (Mac/iPhone/iPad).
Supported audio formats: MP3, M4A/AAC, WAV, CAF, and AIFF.
Includes a new pre-bundled loop sound option ("Mute loop.mp3") available by default.


eerimoq:
I don't quite understand the use case. can you elaborate a bit?

Issue #83, opened 2024-11-21
SRT codec error RELAY server


sanlega:
Hi, I was creating my own SRT relay server in aws and I run into a "problem". I think the problem has to be with the codecs or format of the video. Trying the same resolution/fps on other apps aws receives and relays the signal no problem, but, in moblin, i get a lot of droped packets or something, the image looks gray and only updates some of the areas/time. I'm not really sure if that's the problem or im doing something wrong but I believe it has something to do.

eerimoq:
passphrases have never been tested

mikeful:
What is your SRT latency? Going over public internet using 4G/5G requires at least 1 second and preferably 2.

PR #391, opened 2026-07-01
Add SRT encryption support

### LINKED ISSUE #60: Cloudflare Stream SRT
Hey! Will write up a proper report tomorrow but wanted to quickly note that SRT ingest URLs provided by Cloudflare Stream result in a connect-disconnect loop ad infinitum.

grimfoss:
When SRTLA is disabled (standard SRT), bypass the SRTLA middleware and connect SRT directly to the server over UDP. Fixes streaming to Cloudflare Stream and similar SRT endpoints.

Fixes #60 

Built and verified working on my iPhone 16 Pro Max

### DIFF
--- a/Moblin/Media/Srtla/Client/RemoteConnection.swift
+++ b/Moblin/Media/Srtla/Client/RemoteConnection.swift
@@ -360,7 +360,7 @@ class RemoteConnection: @unchecked Sendable {
             packetsInFlight.insert(getSrtSequenceNumber(packet: packet))
             var numberOfMpegTsPackets = (packet.count - 16) / MpegTsPacket.size
             numberOfNonNullPacketsSent += UInt64(numberOfMpegTsPackets)
-            if numberOfMpegTsPackets < mpegtsPacketsPerPacket {
+            if !isSrtDataPacketEncrypted(packet: packet), numberOfMpegTsPackets < mpegtsPacketsPerPacket {
                 var paddedPacket = packet
                 while numberOfMpegTsPackets < mpegtsPacketsPerPacket {
                     paddedPacket.append(nullPacket)
--- a/Moblin/Media/Srtla/Client/RemoteConnection.swift
+++ b/Moblin/Media/Srtla/Client/RemoteConnection.swift
@@ -77,6 +77,7 @@ class RemoteConnection: @unchecked Sendable {
     private(set) var destinationHost: NWEndpoint.Host?
     private(set) var destinationPort: NWEndpoint.Port?
     private let mpegtsPacketsPerPacket: Int
+    private let packetPadding: Bool
     var typeString: String {
         switch type {
         case .wifi:
@@ -100,6 +101,7 @@ class RemoteConnection: @unchecked Sendable {
     init(
         type: NWInterface.InterfaceType?,
         mpegtsPacketsPerPacket: Int,
+        packetPadding: Bool,
         interface: NWInterface?,
         networkInterfaces: SrtlaNetworkInterfaces,
         priority: Float,
@@ -108,6 +110,7 @@ class RemoteConnection: @unchecked Sendable {
     ) {
         self.type = type
         self.mpegtsPacketsPerPacket = mpegtsPacketsPerPacket
+        self.packetPadding = packetPadding
         self.interface = interface
         self.networkInterfaces = networkInterfaces
         self.priority = priority
@@ -360,7 +363,7 @@ class RemoteConnection: @unchecked Sendable {
             packetsInFlight.insert(getSrtSequenceNumber(packet: packet))
             var numberOfMpegTsPackets = (packet.count - 16) / MpegTsPacket.size
             numberOfNonNullPacketsSent += UInt64(numberOfMpegTsPackets)
-            if !isSrtDataPacketEncrypted(packet: packet), numberOfMpegTsPackets < mpegtsPacketsPerPacket {
+            if packetPadding, numberOfMpegTsPackets < mpegtsPacketsPerPacket {
                 var paddedPacket = packet
                 while numberOfMpegTsPackets < mpegtsPacketsPerPacket {
                     paddedPacket.append(nullPacket)
--- a/Moblin/Media/Srtla/Client/SrtlaClient.swift
+++ b/Moblin/Media/Srtla/Client/SrtlaClient.swift
@@ -42,6 +42,7 @@ class SrtlaClient: NSObject, @unchecked Sendable {
 
     private let networkPathMonitor = NWPathMonitor()
     private let mpegtsPacketsPerPacket: Int
+    private let packetPadding: Bool
     private var host: String = ""
     private var port: Int = 0
     private var groupId: Data?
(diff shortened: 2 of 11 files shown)


eerimoq:
what was the problem?

grimfoss:
Sorry, I meant to add that context to the description.

Moblin gets into a continuous connect/disconnect loop with Cloudflare because the extra hop introduced by the SRTLA middleware interferes with the timing-sensitive SRT encryption handshake, causing the stream to drop immediately after connecting.

This bypass is explicitly restricted to the official implementation because the native Moblin SRT implementation doesn't support passphrase encryption yet. I'm actually working on a separate PR now to add first-class passphrase support for both implementations.

eerimoq:
where can I read more about the "timing-sensitive SRT encryption handshake"?

grimfoss:
I can only speculate as to the exact reason Cloudflare disconnects, but my best guess is that It’s less about an explicit timeout setting and more about how the SRT handshake handles encryption keys across a proxy setup. 

You can check out the details in the official SRT spec (the IETF Internet-Draft) under **[Section 4.3 (Handshake)](https://haivision.github.io/srt-rfc/draft-sharabayko-srt.html#name-handshake-messages)** and **[Section 6 (Encryption)](https://haivision.github.io/srt-rfc/draft-sharabayko-srt.html#encryption))**. Haivision also has a good breakdown in their repo docs under [docs/features/encryption.md](https://github.com/Haivision/srt/blob/master/docs/features/encryption.md).

Basically, during an encrypted handshake, SRT sends Key Material (KM) packets. When Moblin uses the SRTLA tunnel, the path looks like this: `libsrt -> localhost UDP -> SRTLA client -> remote server`. 

Because the local loopback leg responds instantly but the remote leg has real-world latency, strict endpoints like Cloudflare might notice the timing/packet anomalies during the crypto exchange and drop the connection immediately after the initial socket connects. Bypassing the local tunnel and letting `libsrt` talk directly to the remote server via native UDP stops that from happening.

I am making assumptions as to why Cloudflare is rejecting the stream after the connection succeeds, but this change *does* fix the issue. If you need a better answer before you consider merging, I can do more research, but I unfortunately cannot get logs from Cloudflare's end as they do not provide that much control over their streams, so it will still be mainly based on assumptions.

I am also open to looking into just adding proper encryption handling to the SRTLA middleware rather than bypassing it, if that is preferred, although I personally think skipping the middleware when a stream isn't SRTLA makes more sense.

eerimoq:
Yeah i need a good explanation before merging, which most likely results in a different fix.

grimfoss:
I'll see what more I can figure out. I'll see if I can get some info from Cloudflare too.

grimfoss:
I believe I found the actual root cause after adding aggressive logging and tracing the signal flow manually. 

The SRTLA path pads outgoing packets before forwarding them. For unencrypted streams that's harmless, but for encrypted SRT it happens after `libsrt` has already produced the encrypted packet.

Cloudflare accepts the connection, but once encrypted packets start arriving with extra data appended, the endpoint cant decrypt the packets because the packets were modified *after* they were encrypted.

I was incorrectly attributing this to the extra hops that the SRTLA path adds.

I see two immediate solutions to this:

* avoid padding encrypted SRT streams (for example when a `passphrase` is configured), or pad before encryption.
* add proper first-class support for encrypted SRT in the SRTLA path so packet handling remains transparent.

Not padding encrypted streams would be a relatively small fix. Adding first-class encryption support is probably the better long-term solution, since it would also add encryption support to Moblin's SRT implementation. It would also eliminate the need to manually edit the stream URL after creation to restore the passphrase and pbkeylen query parameters, which are currently stripped during the initial URL setup.

What is your preference for me to pursue, and would you rather I close this PR and open a new one when I'm ready, or just update this one?

eerimoq:
That sounds like a more likely problem indeed. You can update this PR

grimfoss:
* added optional passphrase and pbkeylen fields in srt setup that mimic the behavior of the existing streamid field
* skip padding for encrypted packets
* works on official libsrt implementation

Packets arrive at RemoteConnection after they've already been encrypted, so padding is not safe.

I am still working on encryption support for the Moblin implementation, but that is a much more significant project. Given that the goal of the Moblin implementation is to be more efficient, I want to preserve that, which requires a more thoughtful approach than just copying what libsrt is doing. 

Fixes #60 
May also be related to #83 but I have not been able to test this

eerimoq:
I tested streaming encrypted SRT from moblin to ffmpeg without this fix. It works perfectly. So could be a cloudflare bug. Did you contact them?

grimfoss:
I have not gotten a response yet, but reading through the developer discord
it seems that Cloudflare might just be less forgiving of the extra data
padding adds for security and efficiency reasons on their end. I think this
lines up with what happens - the connection succeeds but gets dropped as
soon as video starts flowing.

I haven’t replicated your test with ffmpeg on my end yet but my guess is
ffmpeg can tell which data is garbage and drop just that data. Cloudflare
seemingly doesn’t do this.

I can revert the padding skip and just leave this as having the setup page
not drop the passphrase from the URL. Otherwise, if I get confirmation that
this is the problem, would a better solution just be a padding toggle to
the app?

eerimoq:
I'm not sure what the best way forward is. Maybe we can remove all padding, always. can't remember why I added it.

grimfoss:
Padding appears to have been introduced in `abf9ff8cd` (“SRTLA”) to pad to 6 MPEG-TS packets, then changed in `52beae96c` (“Workaround for irl-srt-server only accepting packets of 1316 bytes”) to pad to 7 packets, then made configurable in `de03a21b2` (“Big SRT packet config”).

So the current 7-packet / 1316-byte behavior appears to have been specifically added as an `irl-srt-server` workaround, but padding itself predates that and was first added with SRTLA.

Current `irlserver/irl-srt-server` still seems to recommend 1316-byte packets, but does not appear to require them anymore. Looking at `irl-srt-server`, that hard requirement seems to have been removed only a few hours after Moblin added the workaround: irl-srt-server commit `359e2c4` happened about four hours after Moblin commit `52beae96c`.

However, since padding predates the irl-srt-server workaround by about a month, I cannot determine the original reason from commit messages alone. Belabox and Moblin’s internal SRTLA server do not appear to have size requirements, and `irl-srt-server` seems to have inherited the 1316-byte requirement from `Edward-Wu/srt-live-server`, which also removed that requirement long before it was removed from `irl-srt-server`.

I can find a clear reason for the later 7-packet behavior, but I cannot find what originally required 6-packet padding, and SLS and irl-srt-server no longer require the padding either.

If padding is removed entirely, would that be split into a separate PR first, then rebase this PR on top of that change?

eerimoq:
let's add a debug toggle to enable/disable adding padding. let's add padding by default to begin with. you can make it part of this PR.

grimfoss:
I added the debug toggle next to the big packets setting, since they are somewhat related.

I also have encryption working in the moblin implementation on another branch. I’m still cleaning it up before submitting, but since it is a larger addition, I wasn’t sure whether you’d prefer that in a separate PR.

eerimoq:
Merged it locally. We had a discussion about the UI changes on my stream, and we decided not to merge that right now. Enter encryption parameters in the URL instead.

eerimoq:
it's in test flight. make sure to test version 33.11.0 (116)

PR #396, opened 2026-07-08
Add altitude ascent/descent tracking for distance and split distance

nofuture2077:
Accumulates GPS altitude gain/loss alongside the existing distance and split distance counters, filtered by location vertical accuracy to reduce noise. Resets follow the same Reset/Split buttons (and !moblin location reset/split chatbot commands) as distance. Adds {altitudeAscent}, {altitudeDescent}, {splitAltitudeAscent} and {splitAltitudeDescent} text widget variables.


nofuture2077:
I will update the algorithm with a more sophisticated approach to have better approximation

Moblin/Various/Model/ModelLocation.swift
```diff
@@ -83,6 +94,7 @@ extension Model {
 
     func updateDistance() {
         let location = locationManager.getLatestKnownLocation()
+        updateAltitude(location: location)
```
eerimoq: what's the benefit of updating altitude from the function that updated distance over having it separately like slope and average speed?

Moblin/Various/Model/ModelLocation.swift
```diff
@@ -1,5 +1,9 @@
 import CoreLocation
 
+// GPS altitude noise is typically several meters even when verticalAccuracy is reported
+// as small, so enforce a floor on top of the reported accuracy.
```
eerimoq: We don't do comments.

Moblin/Various/Model/ModelLocation.swift
```diff
@@ -95,6 +107,32 @@ extension Model {
         }
     }
 
+    // Reference/threshold (hysteresis) algorithm: GPS altitude is noisy, so only commit a
+    // climb/descent once it moves past the reference point by more than the accuracy-based
+    // threshold, then move the reference. Sub-threshold jitter is ignored instead of accumulating.
```
eerimoq: We don't do comments.

Moblin/Various/Model/ModelLocation.swift
```diff
@@ -95,6 +107,32 @@ extension Model {
...
+    private func updateAltitude(location: CLLocation?) {
+        guard let location, location.verticalAccuracy >= 0 else {
```
eerimoq: why include 0?

Moblin/Various/Model/ModelLocation.swift
```diff
@@ -95,6 +107,32 @@ extension Model {
...
+        guard abs(deltaAltitude) >= threshold else {
+            return
+        }
+        if deltaAltitude > 0 {
+            database.location.altitudeAscent += deltaAltitude
+            database.location.splitAltitudeAscent += deltaAltitude
+        } else {
+            database.location.altitudeDescent += -deltaAltitude
+            database.location.splitAltitudeDescent += -deltaAltitude
+        }
```
eerimoq, line 136: I prefer `if delta > threshold {} else if delta < -threshold {}`

Moblin/Various/Model/ModelLocation.swift
```diff
@@ -1,5 +1,9 @@
 import CoreLocation
 
+// GPS altitude noise is typically several meters even when verticalAccuracy is reported
+// as small, so enforce a floor on top of the reported accuracy.
+private let minimumAltitudeThreshold = 3.0
```
eerimoq: I would just inline this one

Moblin/VideoEffects/Text/TextFormatStringLoader.swift
```diff
@@ -314,6 +318,10 @@ class TextFormatLoader {
                     loadItem(part: .debugOverlay, offsetBy: 14)
                 } else if appendSpeedIfPresent(formatFromIndex: formatFromIndex) {
                 } else if appendAverageSpeedIfPresent(formatFromIndex: formatFromIndex) {
+                } else if appendSplitAltitudeAscentIfPresent(formatFromIndex: formatFromIndex) {
+                } else if appendSplitAltitudeDescentIfPresent(formatFromIndex: formatFromIndex) {
+                } else if appendAltitudeAscentIfPresent(formatFromIndex: formatFromIndex) {
+                } else if appendAltitudeDescentIfPresent(formatFromIndex: formatFromIndex) {
```
eerimoq, line 322: move altitude ascent/descent before split ones

eerimoq:
how did you test it?

nofuture2077:
Haven't tested much, yet. I plan to build a local version of moblin and running a bike stream with it showing the numbers. Then compare it to what my wahoo GPS tracker calculates. Will have results by the weekend.

Doing a PR because others might test it as well in a Testflight phase

Moblin/Various/Model/ModelLocation.swift
eerimoq: hmm I changed my mind about this one. original code was better

eerimoq:
I'll wait for your test before merging

Moblin/Various/Model/ModelLocation.swift
```diff
@@ -24,11 +24,18 @@ extension Model {
...
+        database.location.splitAltitudeAscent = 0.0
+        database.location.splitAltitudeDescent = 0.0
         latestKnownLocation = nil
+        altitudeReference = nil
     }
 
     func resetSplitDistance() {
         database.location.splitDistance = 0.0
+        database.location.splitAltitudeAscent = 0.0
+        database.location.splitAltitudeDescent = 0.0
```
eerimoq, line 40: Think about why this is reset here.

Moblin/Various/Model/ModelLocation.swift
eerimoq: Ping.

Moblin/Various/Model/ModelLocation.swift
eerimoq: why did you remove the check all together?

Moblin/Various/Model/ModelLocation.swift
nofuture2077: Think about a multiday bike / hike stream. The split function allows you to have daily distance & ascent. I don't think we need another UI button because splitting ascent without splitting distance seems odd and would always require an additional click.

Moblin/Various/Model/ModelLocation.swift
nofuture2077: i'll revert it then

nofuture2077:
before testing again I asked AI about lowering the threshold. In my opinion moving to the barometer would be the better change - I'm just unsure if it should be in this PR or separate.

Effect of lowering it

The 3.0 m floor is a deliberate noise gate — GPS vertical accuracy is often worse than horizontal, and small readings can easily jitter ±2-5 m even when not moving. Lowering the floor (e.g. to 1-2 m) will:

Capture more of the small real elevation changes (closer to Wahoo Roam's barometric altimeter results), which is likely why you're undercounting (271m vs 301m — about 10% low).
But increase susceptibility to noise-driven false ascent/descent accumulation, especially on flat terrain or standing still — this noise doesn't cancel out because ascent and descent are tracked as separate cumulative sums (every up/down jitter adds to both totals rather than canceling).

Is it advisable to lower it?

The gap you're seeing (271m vs 301m, ~10%) is consistent with known GPS-altitude-based tracking being less sensitive than barometric altimeters (Wahoo Roam uses a barometric sensor, which is more precise for small elevation changes and not subject to this thresholding issue at all). Lowering the threshold from 3.0 to something like 2.0 could help close the gap, but:

If Moblin's device relies on GPS altitude (not barometric), no amount of threshold tuning will fully match a barometric device — you'll always have more noise per meter of accuracy.
A better fix, if the device has a barometer (iPhones since 6 have one, exposed via CMAltimeter), would be to use relative altitude from the barometer instead of CLLocation.altitude, since it's the same sensor class the Wahoo uses and would inherently be more accurate for ascent/descent tracking — that's a bigger change than tuning a constant though.

nofuture2077:
CMAltimeter gives relative altitude change since start (not absolute), only works on real devices (not simulator), and requires a new CoreMotion dependency plus a motion-usage permission string.

So maybe better a separate PR if we want this?

Moblin/Various/Model/ModelLocation.swift
eerimoq: Think more. Didn’t we move some stuff to a separate function earlier? Do the same here. I’m just testing you if you fully rely on AI, or actually try to write good code :) at least I try lol

nofuture2077: I've a degree in CS and working in the industry for 15+ years. If you ask 4 people what "good code" is you get 7 different opinions.

nofuture2077: added another function layer to better match your current style. If this is better code I don't have any opinion

eerimoq:
Yeah let’s leave that for another PR.

eerimoq:
Look into the open comment further up.

Moblin/Various/Model/ModelLocation.swift
nofuture2077: max(location.verticalAccuracy, 3.0) handles the case

nofuture2077:
I won't do a full stream / test with a lower threshold. Looked at the numbers of the altitude while walking inside and they fluctuate on a iPhone 15 Pro on flat ground. I can't guarantee that 3m is the perfect threshold but it seems reasonable and it is AI suggested.

Just answered the last question - hopefully I didn't miss any comment left.

eerimoq:
image

You missed that one.

nofuture2077:
comment was still pending. max(location.verticalAccuracy, 3.0) handles the case Check ">= 0" is not needed

Moblin/Various/Model/ModelLocation.swift
eerimoq: you sure? the docs says altitude value is invalid if verticalAccuracy is zero or negative: https://developer.apple.com/documentation/corelocation/cllocation/verticalaccuracy

eerimoq:
see my latest comment about the vertical accuracy

nofuture2077:
You are right - that's an logic error. Threshold is guarded but I use verticalAccuracy afterwards in calc when it's invalid

Moblin/Various/Model/ModelLocation.swift
```diff
@@ -95,6 +114,29 @@ extension Model {
         }
     }
 
+    func updateAltitude() {
+        guard let location = locationManager.getLatestKnownLocation(), location.verticalAccuracy >= 0 else {
```
eerimoq: Image

Just as I wrote earlier, 0 SHALL NOT be included in the check.

eerimoq:
merged locally

PR #397, opened 2026-07-10
Add telemetry topic to browser widget JS API

nofuture2077:
Lets web overlays subscribe to real-time telemetry (speed, altitude, location, distance, heart rate, weather, workout data) via moblin.subscribe({telemetry: {}}), matching the existing chat and speechToText topics. Gated by the existing "Moblin access" widget setting, delivered once per second alongside text widget updates.

Would update is per once ascent / descent variables are available


nofuture2077:
Wiring up the telemetry over remote control mechanism works in background mode. Will be done in #405

PR #405, opened 2026-07-17
feat: send telemetry over Remote Control streamer connection

nofuture2077:
Adds TelemetryData/RemoteControlEvent.telemetry and wires it into RemoteControlStreamer so we can build custom HUDs

### DIFF
--- a/Moblin/RemoteControl/RemoteControlStreamer.swift
+++ b/Moblin/RemoteControl/RemoteControlStreamer.swift
@@ -50,7 +50,7 @@ protocol RemoteControlStreamerDelegate: AnyObject {
     func remoteControlStreamerTriggerReaction(reaction: RemoteControlReaction)
     func remoteControlStreamerMoveToGimbalPreset(id: UUID)
     func remoteControlStreamerImportSettings(settings: Data, onCompleted: @escaping (Bool) -> Void)
-    func remoteControlStreamerStartStats()
+    func remoteControlStreamerStartStats(filter: RemoteControlStartStatsFilter?)
     func remoteControlStreamerStopStats()
 }
 
@@ -124,7 +124,7 @@ class RemoteControlStreamer {
         send(message: .event(data: .scoreboard(config: config)))
     }
 
-    func sendStatsUpdate(data: Stats) {
+    func sendStatsUpdate(data: RemoteControlStats) {
         guard connected else {
             return
         }
@@ -374,8 +374,8 @@ class RemoteControlStreamer {
             delegate.remoteControlStreamerImportSettings(settings: data) { succeeded in
                 self.send(message: .response(id: id, result: succeeded ? .ok : .error, data: nil))
             }
-        case .startStats:
-            delegate.remoteControlStreamerStartStats()
+        case let .startStats(filter: filter):
+            delegate.remoteControlStreamerStartStats(filter: filter)
             sendEmptyOkResponse(id: id)
         case .stopStats:
             delegate.remoteControlStreamerStopStats()
--- a/Moblin/RemoteControl/RemoteControlStreamer.swift
+++ b/Moblin/RemoteControl/RemoteControlStreamer.swift
@@ -122,6 +122,13 @@ class RemoteControlStreamer {
         send(message: .event(data: .scoreboard(config: config)))
     }
 
+    func sendTelemetryUpdate(data: TelemetryData) {
+        guard connected else {
+            return
+        }
+        send(message: .event(data: .telemetry(data: data)))
+    }
+
     func sendPreview(preview: Data) {
         send(message: .preview(preview: preview))
     }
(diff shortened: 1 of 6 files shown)


nofuture2077:
waiting for the altitude to update

nofuture2077:
Tested this feature with a custom HUD (not finished). 

https://www.twitch.tv/nofuture2077/clip/ZanySilkyStingrayBloodTrail-3YnHrqcdSzlirtzV

eerimoq:
what about using remote scene?

nofuture2077:
I might not fully get it right but it looks to me that RemoteScene feeds data / scenes / widgets into moblin from a assistant. Where this approach pushes data from moblin to the outside. Is this correct?

What would you prefer?

eerimoq:
Ahh true, yeah could be a different use case. Have to think more about it

nofuture2077:
something got wrong on merge. let me fix that

nofuture2077:
What's with this PR? Tested it in multiple streams and works really good. 

One new use case i discovered: I can overlay Maps / Elevationmaps which are computational heavy without the load  Moblin.

eerimoq:
> What's with this PR? Tested it in multiple streams and works really good.

What do you mean by that question?

Moblin/RemoteControl/RemoteControlStreamer.swift
```diff
@@ -122,6 +124,13 @@ class RemoteControlStreamer {
         send(message: .event(data: .scoreboard(config: config)))
     }
 
+    func sendStatsUpdate(data: RemoteControlStats) {
```
eerimoq: rename to sendStats. I'll rename other functions later to follow same pattern

Moblin/Various/Model/Model.swift
```diff
@@ -601,10 +601,13 @@ final class Model: NSObject, ObservableObject, @unchecked Sendable {
     var remoteControlWeb: RemoteControlWeb?
     var isRemoteControlAssistantRequestingPreview = false
     var isRemoteControlAssistantRequestingStatus = false
+    var isRemoteControlAssistantRequestingStats = false
     var isRemoteControlWebRequestingPreview = false
     var remoteControlAssistantRequestingStatusFilter: RemoteControlStartStatusFilter?
+    var remoteControlAssistantRequestingStatsFilter: RemoteControlStartStatsFilter?
     var remoteControlAssistantPreviewUsers: Set<RemoteControlAssistantPreviewUser> = .init()
     var remoteControlAssistantStatusRequested: Bool = false
+    var remoteControlAssistantLatestStats: RemoteControlStats?
```
eerimoq: why does this one exist?

Moblin/Various/Model/Model.swift
```diff
@@ -1222,6 +1225,16 @@ final class Model: NSObject, ObservableObject, @unchecked Sendable {
         isAppActive = notification.name == UIApplication.didBecomeActiveNotification
     }
 
+    private func isStatsFilterFieldEnabled(_ field: KeyPath<RemoteControlStartStatsFilter, Bool?>) -> Bool {
```
eerimoq: let's create one function per filter member.

Moblin/Various/Model/ModelRemoteControl.swift
```diff
@@ -677,6 +733,11 @@ extension Model: @preconcurrency RemoteControlStreamerDelegate {
         makeToast(title: String(localized: "Remote control assistant disconnected"))
         isRemoteControlAssistantRequestingPreview = false
         isRemoteControlAssistantRequestingStatus = false
+        isRemoteControlAssistantRequestingStats = false
+        remoteControlAssistantRequestingStatsFilter = nil
+        startWeatherManager()
+        startGeographyManager()
+        startGForceManager()
```
eerimoq: call remoteControlStreamerStopStats instead?

Moblin/Various/Model/ModelRemoteControl.swift
```diff
@@ -668,6 +719,11 @@ extension Model: @preconcurrency RemoteControlStreamerDelegate {
         makeToast(title: String(localized: "Remote control assistant connected"), subTitle: subTitle)
         isRemoteControlAssistantRequestingPreview = false
         isRemoteControlAssistantRequestingStatus = false
+        isRemoteControlAssistantRequestingStats = false
+        remoteControlAssistantRequestingStatsFilter = nil
+        startWeatherManager()
+        startGeographyManager()
+        startGForceManager()
```
eerimoq: call remoteControlStreamerStopStats instead?

Moblin/Various/Model/Model.swift
```diff
@@ -1222,6 +1224,36 @@ final class Model: NSObject, ObservableObject, @unchecked Sendable {
         isAppActive = notification.name == UIApplication.didBecomeActiveNotification
     }
 
+    private func isGForceStatsFilterEnabled() -> Bool {
+        guard isRemoteControlAssistantRequestingStats else {
+            return false
+        }
+        guard let filter = remoteControlAssistantRequestingStatsFilter else {
```
eerimoq: remove the assignment

eerimoq: or hmm, I think we should make remoteControlAssistantRequestingStatsFilter non-optional, and assign "all filters enabled" to it if we receive it as nil in startStats(). That would simplify the logic here.

eerimoq:
merged it locally

PR #412, opened 2026-07-21
fix: Low Light Boost respects configured FPS as maximum (e.g. 50)

### LINKED ISSUE #409: Low Light Boost forces 60 FPS when configured for 50 FPS
## Description

There appears to be a bug with **Low Light Boost**. When it is enabled, the camera runs at **60 FPS in normal lighting**, even though the frame rate is explicitly set to **50 FPS**.

In low-light conditions, it correctly drops to **30 FPS** as expected.

## Steps to reproduce

1. Set the camera frame rate to **50 FPS**.
2. Enable **Low Light Boost**.
3. Use the camera in normal lighting.
4. Observe that the camera runs at **60 FPS** instead of the configured 50 FPS.
5. Move into low-light conditions and observe that it drops to **30 FPS**.

## Expected behaviour

With the frame rate set to 50 FPS, the camera should remain at **50 FPS in normal lighting** while Low Light Boost is enabled, only lowering the frame rate when required in darker conditions.

## Additional context

We use 50 FPS because it helps reduce flickering from artificial lights running on a 50 Hz power frequency. I noticed the camera had switched back to 60 FPS because the light flickering started happening again.

sm9cc:
Fixes #409

With Low Light Boost on, `setAutoFps()` left min frame duration invalid so the camera could peak at format max (often 60 FPS) in normal light.

Cap `activeVideoMinFrameDuration` to the configured stream FPS so LLB can still lower FPS in the dark but not go above the setting (e.g. 50).

### DIFF
--- a/Moblin/Media/HaishinKit/Media/Video/VideoUnit.swift
+++ b/Moblin/Media/HaishinKit/Media/Video/VideoUnit.swift
@@ -1904,7 +1904,7 @@ final class VideoUnit: NSObject, @unchecked Sendable {
             }
             device.activeColorSpace = colorSpace
             if useAutoFrameRate {
-                device.setAutoFps()
+                device.setAutoFps(maxFrameRate: fps)
                 processor?.delegate.streamSelectedFps(auto: true)
             } else {
                 device.setFps(frameRate: fps)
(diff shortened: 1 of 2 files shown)


eerimoq:
did you test it on multiple devices with various settings? feels like something will crash or not work =)

eerimoq:
looks like a nice fix

sm9cc:
> did you test it on multiple devices with various settings? feels like something will crash or not work =)

I wish I could test it, but I don't own a mac, or have xcode, I think it's simple enough though.

eerimoq:
simple enough.... =)

image

and the docs clearly states that what this PR does is not allowed:

https://developer.apple.com/documentation/avfoundation/avcapturedevice/isautovideoframerateenabled

PR #417, opened 2026-07-24
Guard DockKit behind canImport for simulator builds

RyzingPhoenix:
## Summary
- DockKit has no iOS Simulator SDK, so the unconditional `import DockKit` broke simulator builds with "module not found"
- Wrap the import and DockKit-typed APIs in `Gimbal.swift`, `SettingsGameController.swift`, and `ModelGameController.swift` behind `#if canImport(DockKit)`, with a no-op `Gimbal` stub for the simulator
- Drop the explicit `DockKit.framework` link from `project.pbxproj` since it has no `.tbd` in the simulator SDK either; Swift autolinking still weak-links it for device builds

## Test plan
- [x] Built for iOS Simulator (iPhone 17 Pro) — succeeds, binary has no DockKit reference
- [x] Built for generic iOS device — succeeds, binary still weak-links `DockKit.framework`
- [x] `swiftformat --lint` and `swiftlint --strict` pass on changed files


eerimoq:
Merged it locally.

PR #421, opened 2026-07-30
Added Support for Black Shark Magcool Pro 5

Spillmaker:
Tested both Pro 4 and Pro 5.

Also adjusted the cooling profile a bit so now when the phone no longer need cooling, it will no longer cool. (Previously it kept cooling sligthly)

Linted, so hope this passes the check!

### DIFF
--- a/Moblin/Integrations/BlackSharkCooler/BlackSharkCoolerDevice.swift
+++ b/Moblin/Integrations/BlackSharkCooler/BlackSharkCoolerDevice.swift
@@ -188,43 +188,29 @@ extension BlackSharkCoolerDevice: CBPeripheralDelegate {
         guard let peripheral, let writeCharacteristic, let model else {
             return
         }
-
         let thermalState = ProcessInfo.processInfo.thermalState
-
         switch model {
         case .pro4:
-            // The Pro 4 Operates on  percentage-based values. 0 is off while 100 is max.
             let coolingPowerTarget: Int
             let fanSpeedTarget: Int
-
             switch thermalState {
             case .nominal:
-                // Phone most likely not streaming. We should stop cooling to prevent condensation.
                 coolingPowerTarget = 0
                 fanSpeedTarget = 10
             case .fair:
-                // Our target thermal-state. we can "chill" now. get it haha.
                 coolingPowerTarget = 20
                 fanSpeedTarget = 20
             case .serious:
-                // Risk of reduced performance starts here.
                 coolingPowerTarget = 80
                 fanSpeedTarget = 50
             case .critical:
-                // Phone is overheating! We dont want to be here.
                 coolingPowerTarget = 100
                 fanSpeedTarget = 100
             @unknown default:
-                // We have no idea what this could be. Max everything as a failsafe.
                 coolingPowerTarget = 100
                 fanSpeedTarget = 100
                 logger.info("black-shark-cooler-device: Thermal state is unknown value ( \(thermalState) )")
             }
-            // Since we do not know the fan and cooler-state we have to assume that it can be out of sync.
-            // sending
-            // the commands to update the cooling power and fan speed on every interval will make sure that
-            // its in
-            // sync.
             let coolingPower = updatedPercentageScale(coolingPower, target: coolingPowerTarget)
             logger.debug("black-shark-cooler-device (Pro 4): Adjusting cooling power to \(coolingPower)%")
             peripheral.writeValue(
@@ -242,8 +228,6 @@ extension BlackSharkCoolerDevice: CBPeripheralDelegate {
         case .pro5:
             switch thermalState {
             case .nominal:
-                // Turn cooling off. (AKA Desk mode. Fan will run a bit to remove residual heat. cooling is
-                // fully off)
                 logger.debug("black-shark-cooler-device (Pro 5): Adjusting cooling power to OFF")
                 peripheral.writeValue(
                     BlackSharkLib.getSetCoolingEnabledCommand(false, model: .pro5)!,
@@ -252,19 +236,14 @@ extension BlackSharkCoolerDevice: CBPeripheralDelegate {
                 )
             case .fair:
                 if coolingPower != nil, coolingPower! == 2 {
-                    // Cooling is off, which means we are in the wrong mode. set into custom mode.
                     logger.debug("black-shark-cooler-device (Pro 5): Enabling custom mode for cooler.")
-
                     peripheral.writeValue(
                         BlackSharkLib.getSetCoolingEnabledCommand(true, model: .pro5)!,
                         for: writeCharacteristic,
                         type: .withoutResponse
                     )
                 }
-                // The 5 Pro does not have refined controls for the cooling power and fan.
-                // So here we need to co into the "custom" mode which has 5 steps of intensity.
                 logger.debug("black-shark-cooler-device (Pro 5): Adjusting cooling power to 1")
-
                 peripheral.writeValue(
                     BlackSharkLib.getSetCustomModeCommand(intensity: 1, model: .pro5)!,
                     for: writeCharacteristic,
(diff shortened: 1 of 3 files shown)


Moblin/Integrations/BlackSharkCooler/BlackSharkCoolerDevice.swift
```diff
@@ -170,54 +174,166 @@ extension BlackSharkCoolerDevice: CBPeripheralDelegate {
...
+                        BlackSharkLib.getSetCoolingEnabledCommand(true, model: .pro5)!,
+                        for: writeCharacteristic,
+                        type: .withoutResponse
+                    )
+                }
+                // The 5 Pro does not have refined controls for the cooling power and fan.
+                // So here we need to co into the "custom" mode which has 5 steps of intensity.
+                logger.debug("black-shark-cooler-device (Pro 5): Adjusting cooling power to 3")
+                peripheral.writeValue(
+                    BlackSharkLib.getSetCustomModeCommand(intensity: 2, model: .pro5)!,
```
jeromecornet: Should this be 3 here ? or maybe change the comment above ?

Moblin/Integrations/BlackSharkCooler/BlackSharkCoolerDevice.swift
```diff
@@ -170,54 +174,166 @@ extension BlackSharkCoolerDevice: CBPeripheralDelegate {
...
+            )
+            let fanSpeed = updatedPercentageScale(fanSpeed, target: fanSpeedTarget)
+            logger.debug("black-shark-cooler-device (Pro 4): Adjusting fan speed to \(fanSpeed)%")
+            peripheral.writeValue(
+                BlackSharkLib.getSetFanSpeedCommand(fanSpeed, model: model)!,
+                for: writeCharacteristic,
+                type: .withoutResponse
+            )
+        case .pro5:
+            switch thermalState {
```
jeromecornet: Is there any place for the Pro5 "AI" mode in normal operation rather than controlling the custom power based on temperature ?
It feels like if the phone is already overheating the cooler should probably have been set to a higher cooling mode before reaching a high temperature; and my understanding is that the AI mode adapts to the cooling demand load.
Fan only mode makes sense in .nominal, but maybe in .fair we can use the AI mode, and only use a custom 3, 4 or 5 mode with higher temperatures.
Or maybe 2 different user-selectable cooling ramps (one off-1-2(or3)-5 and the other off-2-4-5) ?

jeromecornet (commented):
Thanks for doing this ! I've been manually controlling my Pro 5 fan so I'm glad this can be automated. 

I had a few questions/comments about the power ramping.

Moblin/Integrations/BlackSharkCooler/BlackSharkCoolerDevice.swift
maarsellow: in my experience the ai mode cools the phone way too much, to a point where it gets nearly too cold to touch. I would rather not use that.

I tested the explained method by turning the cooler on and of manually based on the "flame"-icon. the phone was never cold. I find that perfect, because i dont want any risk of condensation and power consumption as low as possible.

jeromecornet: I've been shooting videos outside in full sun and it does require more cooling than when shooting indoors for sure, but I will test it with this power ramp, if it's not enough I will let you know !

Spillmaker:
Cooling-profiles can probably be a feature request. so that you can customize them in the app if so need be. i dont think it may be neccesary tho.

@eerimoq unless you want me to adjust anything i think this is good to push in.

eerimoq:
remove all blank lines in functions and all comments that are not very important

Moblin/Integrations/BlackSharkCooler/BlackSharkCoolerDevice.swift
```diff
@@ -170,54 +174,133 @@ extension BlackSharkCoolerDevice: CBPeripheralDelegate {
...
-        case .critical:
-            coolingPowerTarget = 100
-            fanSpeedTarget = 100
-        @unknown default:
-            coolingPowerTarget = 100
-            fanSpeedTarget = 100
-            logger.info("black-shark-cooler-device: Thermal state is unknown value")
+    }
+
+    func adjustCoolerProfile() {
```
eerimoq: make this function shorter

Moblin/Integrations/BlackSharkCooler/BlackSharkCoolerDevice.swift
```diff
@@ -170,54 +174,133 @@ extension BlackSharkCoolerDevice: CBPeripheralDelegate {
...
+                    )
+                }
+                logger.debug("black-shark-cooler-device (Pro 5): Adjusting cooling power to 1")
+                peripheral.writeValue(
+                    BlackSharkLib.getSetCustomModeCommand(intensity: 1, model: .pro5)!,
+                    for: writeCharacteristic,
+                    type: .withoutResponse
+                )
+            case .serious:
+                if coolingPower != nil, coolingPower! == 2 {
```
eerimoq: if let coolingPower?

Moblin/Integrations/BlackSharkCooler/BlackSharkCoolerDevice.swift
```diff
@@ -170,54 +174,133 @@ extension BlackSharkCoolerDevice: CBPeripheralDelegate {
...
+                        BlackSharkLib.getSetCoolingEnabledCommand(true, model: .pro5)!,
+                        for: writeCharacteristic,
+                        type: .withoutResponse
+                    )
+                }
+                logger.debug("black-shark-cooler-device (Pro 5): Adjusting cooling power to 5")
+                peripheral.writeValue(
+                    BlackSharkLib.getSetCustomModeCommand(intensity: 5, model: .pro5)!,
+                    for: writeCharacteristic,
+                    type: .withoutResponse
```
eerimoq: Create a function that writes for less duplicated code.

Moblin/Integrations/BlackSharkCooler/BlackSharkCoolerDevice.swift
```diff
@@ -99,14 +101,15 @@ extension BlackSharkCoolerDevice: CBCentralManagerDelegate {
...
-                        advertisementData _: [String: Any],
+                        advertisementData: [String: Any],
                         rssi _: NSNumber)
     {
         guard peripheral.identifier == deviceId else {
             return
         }
         central.stopScan()
         self.peripheral = peripheral
+        advertisedName = advertisementData[CBAdvertisementDataLocalNameKey] as? String
```
eerimoq: maybe calling `model = BlackSharkLib.detectModel(advertisedName: advertisedName)` here instead of later?

Spillmaker:
refactored, but will need some time to run trough tests with the devices again (AKA waiting to push in a new commit until ive done testing)

Spillmaker:
Ok, refactoring is in place

Moblin/Integrations/BlackSharkCooler/BlackSharkCoolerDevice.swift
```diff
@@ -35,6 +35,8 @@ class BlackSharkCoolerDevice: NSObject, @unchecked Sendable {
     private var readCharacteristic: CBCharacteristic?
     private var writeCharacteristic: CBCharacteristic?
     private var latestTransmissionTime = ContinuousClock.now
+    private var model: BlackSharkLib.Model?
+    private var advertisedName: String?
```
eerimoq: there is no need to save this anymore

Spillmaker:
Removed some old guards i had missed as well. model is still required, but advertisedname was indeed not needed in that scope anymore, alltough in the future we may want to add it in to get more prettier names in the bluetooth select list.

Moblin/Integrations/BlackSharkCooler/BlackSharkCoolerDevice.swift
```diff
@@ -170,20 +173,26 @@ extension BlackSharkCoolerDevice: CBPeripheralDelegate {
...
+        }
         let coolingPowerTarget: Int
         let fanSpeedTarget: Int
-        switch ProcessInfo.processInfo.thermalState {
+        switch thermalState {
         case .nominal:
-            coolingPowerTarget = 5
-            fanSpeedTarget = 15
+            coolingPowerTarget = 0
+            fanSpeedTarget = 10
```
eerimoq, line 195: are those changes intentional?

Spillmaker: Yes. Instead of keeping the cooling constantly on, we now turn the cooling off when not needed, and the fan is faintly on just to remove the last bit of surplus heat from the heatsink

PR #422, opened 2026-08-01
Fix video when app is backgrounded

S1nGeN0r:
Found a bug with Enhanced Video Stabilization.
If moblin goes to the background and then comes back, the camera can freeze. The stream keeps running, and audio is fine, but the video never recovers.
I also reproduced the same bug in moblin 33.12 with сore image, so its not related to metalpetal.
metalpetal also had an extra issue where the background frame could become green.

Disable enhanced stabilization while the app is in the background.
Restore it after new camera frames arrive.
Keep the last good frame during background streaming.
Use core Image for the background blur transition to avoid green frames


eerimoq:
not sure about this. looks very complicated

PR #419, opened 2026-07-25
Add BLE cycling speed and cadence (CSC) workout device support

jk1rill:
## Summary
- Add Bluetooth LE Cycling Speed and Cadence (CSC) service `0x1816` support so sensors like Cycplus appear in Workout devices.
- Feed cadence into the existing `{cyclingCadence}` text widget variable (same path as cycling power meters).
- Include unit tests for CSC measurement / cadence calculation.

## Test plan
- [ ] Build Moblin and open Settings → Workout devices
- [ ] Pair a CSC cadence (or speed+cadence) sensor such as Cycplus
- [ ] Confirm device reaches Connected
- [ ] Add `{cyclingCadence}` to a text widget and verify cadence updates while pedaling
- [ ] Confirm existing heart rate / cycling power devices still work

Made with [Cursor](https://cursor.com)

PR #435, opened 2026-09-09
Add BLE cycling speed and cadence (CSC) workout device support

KleMiX:
## Why

Not everyone rides with power meter pedals, but plenty of people have a cheap cadence sensor clipped
to the crank. Those sensors use a Bluetooth profile Moblin does not support, so today they simply do
not appear when you scan — there is no way to get cadence on stream without a power meter. This adds
support for them.

The same profile also covers wheel-mounted speed sensors, so speed comes with it, worked out from
the wheel size you set for that device.

## Summary

Adds Bluetooth LE Cycling Speed and Cadence (CSC, service `0x1816`) support, so cadence-only and
speed+cadence sensors show up under Settings → Workout devices. Today cycling cadence is only
available as a side effect of a power meter (Cycling Power `0x1818`), which means the cheap
standalone sensors — Wahoo RPM, Garmin Cadence 2, Magene, Cycplus, most spin-bike sensors — are
invisible to Moblin: the scanner does not even filter for their service.

- CSC feeds the existing `{cyclingCadence}` variable.
- New `{cyclingSpeed}` variable, with the usual unit options (`{cyclingSpeed:km/h}`, `:mph`, `:m/s`).
- Per-device wheel circumference setting (default 2105 mm), used to derive speed from wheel
  revolutions.

This picks up where #419 left off — thanks @jk1rill for the original work. It is reworked against
the current tree (that PR predates `TextEffectStats` becoming `Variables`) and fixes three gaps:

1. **The wheel circumference was hardcoded.** It is now a per-device setting, applied live while
   connected, so speed is right for something other than a 700×25c wheel.
2. **Cadence could be published twice.** A single peripheral can expose both `0x1818` and `0x1816`,
   and `WorkoutDevice` subscribes to whichever services it discovers, so both parsers wrote
   `cyclingCadence`. The cycling power delegate now passes `cadence: Int?` and yields to CSC when
   CSC is reporting crank revolutions for that peripheral. That is recorded on *field presence*, not
   on a derived value, so the two never both publish while CSC is still waiting for its first delta.
3. **Cadence-only sensors reported a speed of zero.** Speed and cadence are each optional now and
   only reach the model when the sensor actually reports that field.

Robustness details, mirroring what `WorkoutDeviceCyclingPower` already does: 16-bit crank and 32-bit
wheel counter rollover, 1/1024 s event-time base with its own rollover, a zero time delta (duplicate
or retransmitted notification) is skipped rather than treated as a stop, a 3-sample rolling average,
decay to zero after 3 s without new revolution data, and clamps on both the wheel revolution delta
and the derived values.

## Test plan

Tested on an indoor trainer with two separate single-metric CSC sensors — one on the crank, one on
the wheel — plus a power meter and a heart rate strap.

- [x] Settings → Workout devices → add a device, scan, pair a CSC sensor
- [x] Confirm it reaches Connected
- [x] `{cyclingCadence}` in a text widget updates while pedalling and returns to 0 when stopping
- [x] `{cyclingSpeed}` updates from a wheel sensor
- [x] A cadence-only sensor shows cadence and a speed-only sensor shows speed, without either
      overwriting the other
- [x] Existing heart rate and cycling power devices still work alongside them
- [x] With a power meter and a CSC cadence sensor both connected, cadence stays steady

Unit tests: `WorkoutDeviceCyclingSpeedCadenceSuite` covers cadence and speed derivation, both
counter wrap-arounds, cadence-only and speed-only sensors, combined wheel+crank packets, a duplicate
notification with no time progress, and a truncated packet. The wheel-circumference arithmetic is
covered there too, since a trainer gives me nothing to measure derived speed against.

## Screenshots

**Two CSC sensors paired — one on the crank, one on the wheel — each reporting only its own metric** (was hand rolling front wheel)

speed

**Cadence stays steady with a power meter and a heart rate strap connected alongside**

BLE

**The setup: an indoor trainer, which is why the speed and distance figures are low**

bike

## Attribution

an outdoor bike.

The Bluetooth parsing was additionally cross-checked against a working Cycling Speed and Cadence
implementation in my own [Virtual Cycling Trainer](https://apps.apple.com/us/app/virtual-cycling-trainer/id6752381425) app, which has been running against real sensors
for a while — the flag handling, the 1/1024 s event-time base, the counter rollover and the
stopped-versus-no-new-event distinction all match what that stack does.

### DIFF
--- a/Moblin/Integrations/WorkoutDevice/WorkoutDevice.swift
+++ b/Moblin/Integrations/WorkoutDevice/WorkoutDevice.swift
@@ -14,8 +14,7 @@ protocol WorkoutDeviceDelegate: AnyObject {
     func workoutDeviceState(_ device: WorkoutDevice, state: WorkoutDeviceState)
     func workoutDeviceHeartRate(_ device: WorkoutDevice, heartRate: Int)
     func workoutDeviceCyclingPower(_ device: WorkoutDevice, power: Int, cadence: Int?)
-    func workoutDeviceCyclingSpeedCadence(_ device: WorkoutDevice,
-                                          metrics: WorkoutDeviceCyclingSpeedCadenceMetrics)
+    func workoutDeviceCyclingSpeedCadence(_ device: WorkoutDevice, speed: Double?, cadence: Int?)
     func workoutDeviceRunningMetrics(_ device: WorkoutDevice, metrics: WorkoutDeviceRunningMetrics)
 }
 
@@ -32,14 +31,18 @@ class WorkoutDevice: NSObject, @unchecked Sendable {
     private var peripheral: CBPeripheral?
     private let heartRate = WorkoutDeviceHeartRate()
     private let cyclingPower = WorkoutDeviceCyclingPower()
-    private let cyclingSpeedCadence = WorkoutDeviceCyclingSpeedCadence()
+    private let cyclingSpeedCadence: WorkoutDeviceCyclingSpeedCadence
     private let running = WorkoutDeviceRunning()
     private var deviceId: UUID?
     weak var delegate: (any WorkoutDeviceDelegate)?
 
-    func start(deviceId: UUID?, wheelCircumference: Int) {
+    init(wheelCircumference: Int) {
+        cyclingSpeedCadence = WorkoutDeviceCyclingSpeedCadence(wheelCircumference: wheelCircumference)
+        super.init()
+    }
+
+    func start(deviceId: UUID?) {
         dispatchQueue.async {
-            self.cyclingSpeedCadence.setWheelCircumference(millimeters: wheelCircumference)
             self.startInternal(deviceId: deviceId)
         }
     }
@@ -155,15 +158,15 @@ extension WorkoutDevice: CBCentralManagerDelegate {
 
     private func handleCyclingPowerMeasurement(value: Data) throws {
         let (power, cadence) = try cyclingPower.handleMeasurement(value: value)
-        let ownedByCyclingSpeedCadence = cyclingSpeedCadence.isReportingCrankRevolutions()
+        let ownedByCyclingSpeedCadence = cyclingSpeedCadence.isReportingCadence()
         delegate?.workoutDeviceCyclingPower(self,
                                             power: power,
                                             cadence: ownedByCyclingSpeedCadence ? nil : cadence)
     }
 
     private func handleCyclingSpeedCadenceMeasurement(value: Data) throws {
-        let metrics = try cyclingSpeedCadence.handleMeasurement(value: value)
-        delegate?.workoutDeviceCyclingSpeedCadence(self, metrics: metrics)
+        let (speed, cadence) = try cyclingSpeedCadence.handleMeasurement(value: value)
+        delegate?.workoutDeviceCyclingSpeedCadence(self, speed: speed, cadence: cadence)
     }
 
     private func handleCyclingPowerVector(value: Data) throws {
--- a/Moblin/Integrations/WorkoutDevice/WorkoutDevice.swift
+++ b/Moblin/Integrations/WorkoutDevice/WorkoutDevice.swift
@@ -37,9 +37,9 @@ class WorkoutDevice: NSObject, @unchecked Sendable {
     private var deviceId: UUID?
     weak var delegate: (any WorkoutDeviceDelegate)?
 
-    func start(deviceId: UUID?, wheelCircumferenceMillimeters: Int) {
+    func start(deviceId: UUID?, wheelCircumference: Int) {
         dispatchQueue.async {
-            self.cyclingSpeedCadence.setWheelCircumference(millimeters: wheelCircumferenceMillimeters)
+            self.cyclingSpeedCadence.setWheelCircumference(millimeters: wheelCircumference)
             self.startInternal(deviceId: deviceId)
         }
     }
@@ -58,10 +58,6 @@ class WorkoutDevice: NSObject, @unchecked Sendable {
 
     func getState() -> WorkoutDeviceState {
         state
-    }
-
-    func isCyclingSpeedCadenceDiscovered() -> Bool {
-        cyclingSpeedCadence.isAnyCharacteristicDiscovered()
     }
 
     private func startInternal(deviceId: UUID?) {
--- a/Moblin/Integrations/WorkoutDevice/WorkoutDevice.swift
+++ b/Moblin/Integrations/WorkoutDevice/WorkoutDevice.swift
@@ -6,13 +6,16 @@ private let dispatchQueue = DispatchQueue(label: "com.eerimoq.workout-device")
 nonisolated(unsafe) let workoutDeviceScanner = BluetoothScanner(serviceIds: [
     workoutDeviceHeartRateServiceId,
     workoutDeviceCyclingPowerServiceId,
+    workoutDeviceCyclingSpeedCadenceServiceId,
     workoutDeviceRunningServiceId,
 ])
 
 protocol WorkoutDeviceDelegate: AnyObject {
     func workoutDeviceState(_ device: WorkoutDevice, state: WorkoutDeviceState)
     func workoutDeviceHeartRate(_ device: WorkoutDevice, heartRate: Int)
-    func workoutDeviceCyclingPower(_ device: WorkoutDevice, power: Int, cadence: Int)
+    func workoutDeviceCyclingPower(_ device: WorkoutDevice, power: Int, cadence: Int?)
+    func workoutDeviceCyclingSpeedCadence(_ device: WorkoutDevice,
+                                          metrics: WorkoutDeviceCyclingSpeedCadenceMetrics)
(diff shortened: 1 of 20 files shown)


Moblin/Various/Model/ModelWorkoutDevice.swift
```diff
@@ -31,6 +37,10 @@ extension Model {
         workoutDevices[device.id]?.getState() ?? .disconnected
     }
 
+    func isWorkoutDeviceCyclingSpeedCadence(device: SettingsWorkoutDevice) -> Bool {
+        workoutDevices[device.id]?.isCyclingSpeedCadenceDiscovered() ?? false
+    }
+
```
eerimoq: I don't like this as WorkoutDevice.cyclingSpeedCadence is kinda owned by another dispatch queue. I suggest removing the function and always show wheel size in settings

Moblin/Integrations/WorkoutDevice/WorkoutDevice.swift
```diff
@@ -140,7 +159,15 @@ extension WorkoutDevice: CBCentralManagerDelegate {
 
     private func handleCyclingPowerMeasurement(value: Data) throws {
         let (power, cadence) = try cyclingPower.handleMeasurement(value: value)
-        delegate?.workoutDeviceCyclingPower(self, power: power, cadence: cadence)
+        let ownedByCyclingSpeedCadence = cyclingSpeedCadence.isReportingCrankRevolutions()
```
eerimoq: why is cyclingSpeedCadence prioritized?

Moblin/Integrations/WorkoutDevice/WorkoutDeviceCyclingSpeedCadence.swift
```diff
@@ -0,0 +1,180 @@
...
+        let now = ContinuousClock.now
+        updateCadence(measurement: measurement, now: now)
+        updateSpeed(measurement: measurement, now: now)
+        return WorkoutDeviceCyclingSpeedCadenceMetrics(
+            speed: reportsWheelRevolutions ? averageSpeed.averageIgnoreZeros() : nil,
+            cadence: reportsCrankRevolutions ? Int(averageCadence.averageIgnoreZeros()) : nil
+        )
+    }
+
+    private func updateCadence(measurement: CyclingSpeedCadenceMeasurement, now: ContinuousClock.Instant) {
```
eerimoq: this function is mostly identical to the one in WorkoutDeviceCyclingPower. is there a good reason for it, or just to keep PR diff minimal?

Moblin/Various/Settings/Settings.swift
```diff
@@ -669,6 +672,9 @@ class SettingsWorkoutDevice: Codable, Identifiable, ObservableObject, Named {
         enabled = container.decode(.enabled, Bool.self, false)
         bluetoothPeripheralName = try? container.decode(String.self, forKey: .bluetoothPeripheralName)
         bluetoothPeripheralId = try? container.decode(UUID.self, forKey: .bluetoothPeripheralId)
+        wheelCircumferenceMillimeters = container.decode(.wheelCircumferenceMillimeters,
+                                                         Int.self,
+                                                         defaultWheelCircumferenceMillimeters)
```
eerimoq: We typically don't have units in names in settings

Moblin/Various/Model/ModelWorkoutDevice.swift
KleMiX: Good point, removed it. Wheel circumference is now always shown.

Moblin/Integrations/WorkoutDevice/WorkoutDevice.swift
KleMiX: Some trainers expose both services and send crank data in both, so without picking one the cadence jumps between two values. CSC is the dedicated cadence service, so I went with that. No strong preference though, happy to prefer power instead.

Moblin/Integrations/WorkoutDevice/WorkoutDeviceCyclingSpeedCadence.swift
KleMiX: Mostly to keep the diff small. Moved it into a shared helper used by both.

Moblin/Various/Settings/Settings.swift
KleMiX: Renamed to wheelCircumference.

Moblin/Integrations/WorkoutDevice/WorkoutDevice.swift
```diff
@@ -140,7 +155,15 @@ extension WorkoutDevice: CBCentralManagerDelegate {
...
```
eerimoq: what about renaming it to isReportingCadence?

Moblin/Integrations/WorkoutDevice/WorkoutDeviceCyclingSpeedCadence.swift
```diff
@@ -0,0 +1,122 @@
...
+                                          time: measurement.lastCrankEventTime,
+                                          now: now)
+        updateSpeed(measurement: measurement, now: now)
+        return WorkoutDeviceCyclingSpeedCadenceMetrics(
+            speed: reportsWheelRevolutions ? averageSpeed.averageIgnoreZeros() : nil,
+            cadence: reportsCrankRevolutions ? cadence : nil
+        )
+    }
+
+    private func updateSpeed(measurement: CyclingSpeedCadenceMeasurement, now: ContinuousClock.Instant) {
```
eerimoq, line 81: It looks like WorkoutDeviceCyclingPower also has cumulativeWheelRevolutions and lastWheelEventTime, so should be calculate speed for that BLE device service as well? Can you test it, cause I can't?

Moblin/Integrations/WorkoutDevice/WorkoutDeviceCyclingSpeedCadence.swift
```diff
@@ -0,0 +1,122 @@
...
+
+let defaultWheelCircumference = 2105
+
+private let measurementWheelRevolutionDataFlagIndex = 0
+private let measurementCrankRevolutionDataFlagIndex = 1
+
+struct WorkoutDeviceCyclingSpeedCadenceMetrics {
+    var speed: Double?
+    var cadence: Int?
+}
```
eerimoq: Remove this and do as cycling power does. I know running is different, but let's do like this for now.

Moblin/Integrations/WorkoutDevice/WorkoutDeviceCyclingSpeedCadence.swift
```diff
@@ -0,0 +1,122 @@
+@preconcurrency import CoreBluetooth
+import Foundation
+
+let workoutDeviceCyclingSpeedCadenceServiceId = CBUUID(string: "1816")
+let workoutDeviceCyclingSpeedCadenceMeasurementCharacteristicId = CBUUID(string: "2A5B")
+
+let defaultWheelCircumference = 2105
```
eerimoq: move this to settings, and pass it to init() of WorkoutDevice and children. Should not be too messy I hope. So no default value of WorkoutDeviceCyclingSpeedCadence.wheelCircumferenceMeters.

Moblin/Integrations/WorkoutDevice/WorkoutDeviceCyclingSpeedCadence.swift
KleMiX: Done, it returns a tuple now like cycling power.

Moblin/Integrations/WorkoutDevice/WorkoutDeviceCyclingSpeedCadence.swift
KleMiX: Done. The default is in settings now and the wheel circumference is passed to init() of WorkoutDevice and WorkoutDeviceCyclingSpeedCadence. Changing it in settings still updates it live.

Moblin/Integrations/WorkoutDevice/WorkoutDevice.swift
KleMiX: Done.

Moblin/Integrations/WorkoutDevice/WorkoutDeviceCyclingSpeedCadence.swift
KleMiX: I checked my trainer (Van Rysel D100). It lists wheel revolution data as supported, but its Cycling Power measurements only contain power and pedal balance. My Assioma pedals don't send it either, so I have no real device for it.

I have my own trainer simulator app that acts as a real BLE device, so I could add wheel data to its Cycling Power packets and test against that. It wouldn't be real hardware though. Would that be good enough, or should we leave speed from cycling power out for now?

eerimoq: Let's skip it then

PR #436, opened 2026-09-09
Cycling cadence from Apple Watch workouts

KleMiX:
## Why

If you start your workouts from the Apple Watch, your sensors are usually paired to the watch rather
than the phone. The watch is already reading cadence from them — it just never passes it to the
phone, so the text widget shows nothing. This sends it across so it can go on stream.

## Summary

A workout started on the Apple Watch never delivers cycling cadence to the stream, even when a
Bluetooth cadence sensor is paired to the watch. Cycling power has the same problem. watchOS 10 added pairing
cycling sensors directly to the watch along with the `cyclingPower` / `cyclingCadence` HealthKit
types, and the watch app's deployment target is already watchOS 10.2 — the data is simply dropped.

Two separate problems:

- **`cyclingPower` is requested but discarded.** It is in `healthKitTypes` in `WatchModel.swift`, so
  `HKLiveWorkoutDataSource` collects it, but `WatchProtocolWorkoutStats.init(statistics:)` has no
  case for it and falls through to `default: break`. The watch then sends an all-`nil` stats message
  to the phone for every sample, which `Model.handleWorkout` ignores because every field is
  `if let`-guarded. So it costs a WatchConnectivity message per sample and produces nothing.
- **`cyclingCadence` is never requested at all** — it appears nowhere in the repo.

This adds both to the wire format and maps them, so `{cyclingPower}` and `{cyclingCadence}` work
from a watch workout the same way they already do from a phone-paired sensor.

## Notes for review

- `WatchProtocol.swift` is shared with the iOS target, whose deployment target is 16.4, while
  `cyclingPower` / `cyclingCadence` need iOS 17. Rather than putting `@available` on `switch` cases
  over `statistics.quantityType`, the two are handled in an `if #available(iOS 17.0, *)` block that
  returns early, before the existing switch. This is the same guard `ModelWorkout.swift` already
  uses around its `types()` set.
- `WatchProtocolWorkoutStats` gains two optional fields. It is `Codable`, so an older phone paired
  with a newer watch (or the reverse) still decodes — new optionals simply come back `nil`.
- Collection is enabled explicitly for both types on a cycling workout, mirroring how `stepCount` is
  already enabled for walking and running. `HKLiveWorkoutDataSource` collects them by default on
  watchOS 10+, so this is belt and braces.
- **Precedence.** `Model.cyclingPower` and `Model.cyclingCadence` are single values shared with the
  phone's own Bluetooth workout devices, so a watch-paired and a phone-paired sensor would otherwise
  fight. Watch-sourced values are applied only when no workout device has published cycling data in
  the last 5 seconds, which keeps the lower-latency direct connection authoritative without
  disabling the watch path for someone who has, say, only a heart rate strap paired to the phone.
- iOS-started workouts (`ModelWorkout.swift`, iOS 26+) get the same two types, so a phone workout
  with a watch-paired sensor benefits too.

## Known missing: cycling speed

`cyclingSpeed` is not covered here. It is not in the watch's read set, has no field on
`WatchProtocolWorkoutStats`, and nothing maps it — so a wheel sensor paired to the watch still
produces nothing, while its cadence now comes through.

It is deliberately left out rather than overlooked. The `{cyclingSpeed}` variable and the
`Model.cyclingSpeed` it reads are introduced by the cycling speed and cadence (CSC) sensor change,
not by this one, so adding it here would either make this PR depend on that one or duplicate its
plumbing and guarantee a conflict between them. It also needs the same precedence rule against the
CSC-derived value that cadence has. Better as a small follow-up once both have landed: the read
type, a field, a mapping and the guard, roughly ten lines.

## Test plan

- [x] Pair a cycling cadence sensor to the Apple Watch
- [x] Start a cycling workout from the watch app
- [x] `{cyclingCadence}` in a text widget updates while pedalling
- [x] Confirmed the value comes from the watch — all Bluetooth workout devices were deleted from the
      phone first, so there was no other cadence source

Cadence is what the screenshots show. The power mapping is the same three lines applied to
`cyclingPower`, which is already in the read set upstream and simply never mapped — say the word and
I will drop that half and keep this to cadence alone.

## Screenshots

**Cadence coming from the watch — the phone has no Bluetooth workout devices configured at all**

watch

**The watch running the workout, the phone showing the value on the overlay**

watch data

## Attribution

an outdoor bike.


Moblin Watch/Shared/WatchProtocol.swift
```diff
@@ -141,8 +141,26 @@ struct WatchProtocolWorkoutStats: Codable {
...
+            }
+            if statistics.quantityType == HKQuantityType.quantityType(forIdentifier: .cyclingCadence) {
+                if let cyclingCadence = statistics.mostRecentQuantity()?
+                    .doubleValue(for: .count().unitDivided(by: HKUnit.minute()))
+                {
+                    self.cyclingCadence = Int(cyclingCadence)
+                }
+                return
+            }
+        }
```
eerimoq: make it a switch statement, and move it after the switch below

eerimoq: And when you are at it, we should probably not send one message per workout stat type, but send one message that contains all stats.

so for loop in HKLiveWorkoutBuilderDelegate.workoutBuilder() should build the message in the loop, and send it after

Moblin/Various/Model/Model.swift
```diff
@@ -671,6 +671,8 @@ final class Model: NSObject, ObservableObject, @unchecked Sendable {
     var catPrinters: [UUID: CatPrinter] = [:]
     var cyclingPower = 0
     var cyclingCadence = 0
+    var latestWorkoutDeviceCyclingPowerTime: ContinuousClock.Instant?
+    var latestWorkoutDeviceCyclingCadenceTime: ContinuousClock.Instant?
     var latestCyclingSpeedCadenceCadenceTime: ContinuousClock.Instant?
```
eerimoq: one latest time variable per source should be enough, now there are three

Moblin Watch/Shared/WatchProtocol.swift
KleMiX: Done.

KleMiX: Done. init(statistics:) is now update(statistics:), the loop fills one stats object and sends it once after the loop. Nothing is sent if there's nothing in it. Did the same for the identical loop on the phone side.

Moblin/Various/Model/Model.swift
KleMiX: Down to two, but it needs a bit more than a plain timestamp, so each one now records which source set it: latestCyclingPower and latestCyclingCadence.

A value is taken if its source ranks at least as high as the last one, or if the last update is older than 5 seconds. Ranking is CSC sensor > power meter > watch.

One timestamp per source isn't quite enough, because a source doesn't always carry both values. A speed-only CSC sensor sends no cadence, and a power meter without crank data sends no cadence either, so a single "this source was heard from" timestamp would block the watch's cadence even though nothing else is providing it. Keeping it per value instead of per source also makes the CSC-over-power-meter rule fall out of the same comparison.

Moblin/Various/Model/ModelWorkoutDevice.swift
```diff
@@ -57,11 +63,29 @@ extension Model {
...
+    func setCyclingCadence(_ cadence: Int, source: CyclingSource) {
+        guard isBetterCyclingSource(latest: latestCyclingCadence, source: source) else {
+            return
         }
-        return latestCyclingSpeedCadenceCadenceTime.duration(to: .now) < .seconds(5)
+        cyclingCadence = cadence
+        latestCyclingCadence = (source, .now)
+    }
+
+    private func isBetterCyclingSource(latest: (source: CyclingSource, time: ContinuousClock.Instant)?,
```
eerimoq: I think we should make this a method on the CyclingSource enum

KleMiX: Done.

PR #437, opened 2026-09-09
Store workout device data in workouts started on the iPhone

KleMiX:
## Why

Sensors paired directly in Moblin are only ever shown on screen. Nothing they measure is written
into the workout, so the workout you save is missing readings Moblin had the whole time. This stores
them, so the ride you recorded matches what you were looking at.

## Summary

A workout started on the iPhone records only what HealthKit itself can collect. The Bluetooth
workout devices Moblin already talks to — heart rate straps, cycling power meters, CSC sensors — are
display-only: their values reach the text widget and then vanish.

The gap is not that the phone cannot measure these things. iOS 26 happily takes heart rate from a
system-paired strap or from AirPods Pro 3. The gap is that Moblin connects to its workout devices
directly over CoreBluetooth, so a sensor bonded to Moblin is not bonded to iOS and HealthKit never
sees it — and since most of these sensors accept only one connection, pairing the same sensor to
both is not an option. Cycling power and cadence have no system-level pairing on iPhone at all.

This feeds the values Moblin already has into the live workout builder, so the saved workout
contains them.

## Depends on

`workout-end-collection` (PR 0) must land first. It makes the workout ending sequence follow
Apple's documented order, without which saving is unreliable and there is nothing to observe these
samples in.

## Notes for review

- Samples are only added while `workoutSession?.state == .running`, so nothing is written unless the
  phone itself owns the session. A workout running on the Apple Watch is untouched.
- Throttled to one sample per second per metric. Cycling power notifications can arrive at 4 Hz,
  which would be ~14k samples an hour for one metric.
- `typesToShare` grows from just `workoutType()` to also include heart rate and, behind
  `#available(iOS 17.0, *)`, cycling power and cadence — writing a sample requires share
  authorization for its type, so this widens the permission prompt.
- The whole `Workout` class is already `@available(iOS 26.0, *)`, so this only benefits iOS 26+.

## Open question

Heart rate can double up. If the user has another heart rate source already feeding HealthKit — a
system-paired strap, AirPods Pro 3, or a paired Apple Watch — then `HKLiveWorkoutDataSource` is
collecting heart rate for the same workout, and the samples added here become a second series.

Cycling power and cadence do not have this problem, since there is no system-level pairing for them
on iPhone.

Options, and I do not have a strong opinion: add heart rate only when the builder is not already
collecting it, add it unconditionally and accept the overlap, or drop heart rate from this change
and keep it to power and cadence. Happy to take direction here.

Running metrics from RSC sensors (`speed`, `cadence`, `distance`) are deliberately left out: distance
would overlap with what the workout builder already accumulates, and getting that wrong corrupts the
saved workout rather than just adding noise.

## Test plan

- [x] On iOS 26+, pair a heart rate strap under Settings → Workout devices
- [x] Start a workout from the Workout quick button (grant the Health permissions when asked)
- [x] Ride for a few minutes, then end the workout
- [x] Open the Health app → the saved workout shows heart rate
- [x] Cycling power and cadence appear in the saved workout as well
- [x] Start a workout from the Apple Watch instead and confirm nothing changed there

Ridden on an indoor trainer with a heart rate strap, a power meter and CSC cadence and speed
sensors connected as Moblin workout devices. The values shown live on the overlay during the ride
match what the saved workout ended up with — heart rate, power and cadence all present.

Depends on the `endCollection` fix; without it the workout is never saved and none of this is
observable.

## Screenshots

**Live on the overlay during the ride — heart rate, cadence and power from Moblin's own sensors**

BLE

**The same values in the workout that was saved**

iOS WRK

## Attribution

an outdoor bike.


KleMiX:
Conflicts fixed.

Moblin/Various/Model/ModelWorkout.swift
```diff
@@ -25,13 +25,16 @@ private class Workout: NSObject {
...
+    private var latestSampleTimes: [HKQuantityTypeIdentifier: ContinuousClock.Instant] = [:]
+    private var workoutType: WatchProtocolWorkoutType?
 
     func isActive() -> Bool {
         workoutSession != nil
     }
 
     func start(model: Model, type: WatchProtocolWorkoutType) -> Bool {
         self.model = model
+        workoutType = type
```
eerimoq: why isn't this assignment in non-macCatalyst branch below?

Moblin/Various/Model/ModelWorkout.swift
```diff
@@ -224,10 +271,36 @@ extension Model {
...
+    func addWorkoutCyclingCadence(_ cadence: Int) {
+        guard #available(iOS 26, *) else {
+            return
+        }
+        Workout.shared.add(cyclingCadence: cadence)
+    }
+
     private func authorizeHealthKit(completion: @escaping @MainActor () -> Void) {
-        let typesToShare: Set = [
+        var typesToShare: Set<HKSampleType> = [
```
eerimoq: why adding the type?

Moblin/Various/Model/ModelWorkout.swift
```diff
@@ -142,8 +146,51 @@ private class Workout: NSObject {
...
+    func add(cyclingCadence: Int) {
+        guard workoutType == .cycling else {
+            return
+        }
+        add(identifier: .cyclingCadence,
+            unit: .count().unitDivided(by: .minute()),
+            value: Double(cyclingCadence))
+    }
+
+    private func add(identifier: HKQuantityTypeIdentifier, unit: HKUnit, value: Double) {
```
eerimoq: would replacing HKQuantityTypeIdentifier with HKQuantityType work? would remove lives 179-181. is there a drawback?

Moblin/Various/Model/ModelWorkout.swift
KleMiX: Moved it into the non-Catalyst branch.

Moblin/Various/Model/ModelWorkout.swift
KleMiX: Removed.

Moblin/Various/Model/ModelWorkout.swift
KleMiX: Yes, those lines are removed now.

PR #445, opened 2026-09-17
Chat platform fixes: Twitch /me and IRC parsing, chat bot quotes, Kick and Soop

gabbelitoV2:
Small independent fixes in the chat platform code, one commit each.

- **Twitch `/me` messages**: `\u{01}ACTION` was stripped with `dropFirst(7)`, leaving the leading space and the trailing `\u{01}`. Emote ranges in the `emotes=` tag are relative to the bare text, so every emote in an action message was cut one scalar too early, and the control character ended up in the text (TTS, cat printer).
- **IRC line splitting**: `components(separatedBy: .whitespaces)` also splits on U+3000 (ideographic space), U+00A0 and other Unicode spaces. A tag value such as `reply-parent-msg-body` containing one broke the tag block apart, the command failed to parse and the whole PRIVMSG was dropped. IRC only separates on ASCII space. Test added.
- **Twitch JOIN**: the channel name is sent as typed. Twitch IRC channels are lowercase logins, so a manually entered name with capitals joined nothing while the chat showed connected.
- **Chat bot quotes**: `popFirst()` only looked for the closing quote on later words, so `!moblin widget "Foo" enable` consumed the rest of the command and returned nil. Test added.
- **Soop chat**: `stop()` did not cancel the web socket (unlike OpenStreamingPlatformChat), so every reload left the old connection receiving until the server dropped it.
- **Kick gifted subscriptions toast** was gated on `kickChatAlerts` instead of `kickToastAlerts` like the other Kick toasts.
- **Kick default categories** were fetched with the model's current stream token instead of the stream being edited, unlike the search right below it.
- **Kick raid channel list** reset `channels` but not `cursor` on appear, so reopening it resumed from the old page.


gabbelitoV2:
Split into one PR per fix: #447, #448, #449, #450, #451, #452, #453, #454.

PR #446, opened 2026-09-17
Settings fixes: lock screen quick button, export failure, stream duplication

gabbelitoV2:
Three small persistence fixes in Settings, one commit each.

- **Lock screen quick button**: `addMissingQuickButtonsPageThree` builds the `.lockScreen` button but the `updateQuickButton` call was lost in fe1308eff, so databases created or reset since then have no lock screen button and existing ones never get its name/icons refreshed.
- **Settings export**: on a failed zip write the `catch` reported `nil` but then fell through and also reported the URL, so the share sheet got a partial or nonexistent file.
- **Duplicating a stream** dropped `youTubeScheduleStreamTitle`/`Visibility`/`AutoStop`, `adaptiveEncoderResolutionThreashold` and `backgroundStreamingPiP`, the recording override (`overrideStream`, `resolution`, `fps`), and `implemenationMigrated` on the SRT settings. The last one made the clone re-run the one-time SRT implementation migration on the next start, switching a Moblin-implementation stream with latency below 1000 ms to the official implementation.


gabbelitoV2:
Split into one PR per fix: #455, #456, #457.

PR #447, opened 2026-09-17
Strip the whole /me wrapper from Twitch action messages

gabbelitoV2:
Twitch sends `/me` messages as `\u{01}ACTION <text>\u{01}`. `dropFirst(7)` only removed `\u{01}ACTION`, leaving the leading space and the trailing control character. The `emotes=` tag offsets are relative to the bare text, so every emote in an action message was cut one scalar too early, and the `\u{01}` ended up in the message text (TTS, cat printer, copy).


eerimoq:
Merged locally.

PR #452, opened 2026-09-17
Use the toast alert setting for the Kick gifted subscriptions toast

gabbelitoV2:
The toast was gated on `kickChatAlerts.giftedSubscriptions` instead of `kickToastAlerts.giftedSubscriptions` like the other Kick toasts, so the toast setting had no effect and the chat setting silenced both.


gabbelitoV2:
Already fixed upstream in 73d6bf2bb (Fix kick toast setting).

PR #490, opened 2026-09-17
Mask PES timestamp bits above 33 when encoding

gabbelitoV2:
PTS/DTS are derived from the host clock, so after about 26.5 hours of device uptime bit 33 and above are set. `encode` shifted the value into the first byte without masking, polluting the prefix bits that must be 0010/0011/0001. Demuxers generally mask it, but it is a spec violation on every stream from a phone that has been up for more than a day.


eerimoq:
Merged locally.

